PR #1, opened 2023-03-22
Add Dockerfile

boomanaiden154:
This commit adds a Dockerfile to the repository containing all of the dependencies necessary to run/develop Gematria. We have found this pretty useful for the [ml-compiler-opt](https://github.com/google/ml-compileropt) repository as it allows for new users to quickly experiment without having to deal with "works on my machine errors" assuming they already have Docker somewhat integrated into their workflow (in addition to having some consumers using containers in production)

I typically use container based development workflows, and I think this might be helpful for others as well (hence why I've opened a PR) while also not adding much maintenance burden. If it's preferred that this isn't viewed as officially supported, we could do something similar to ml-compiler-opt where the `Dockerfile` is in an experimental folder with a note mentioning it is not officially supported.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM ubuntu:22.04
+RUN apt-get update && apt-get install -y gcc python3 python3-pip git curl
+ARG bazelisk_version=1.16.0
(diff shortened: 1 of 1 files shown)


ondrasej:
Thanks!

I'm planning to add a similar Docker config structure as in https://github.com/google/or-tools/tree/stable/tools/docker. But it might take some time before I get to it, and it's great to have a Dockerfile that works until that is done.

PR #2, opened 2023-05-28
Fix Docker Build

boomanaiden154:
Without this patch, the docker build fails in something related to local_config_cc (not really fluent enough in bazel to know more than that it's local c/c++ toolchain related). Switching to clang fixes the issue. Not quite sure why that is (maybe some new project explicitly requests clang?)

This patch also updates Bazelisk in the container image to the latest release.


ondrasej:
Thanks for fixing this!

PR #3, opened 2023-05-28
Remove requirements.txt file from repository

boomanaiden154:
Having this file checked into the repository means that the ignore line in .gitignore is essentially nonfunctional. This patch removes requirements.txt so that the ignore line in .gitignore is functional and the generated requirements.txt doesn't accidentally end up in any commits.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,0 @@
-# Run `bazel run :requirements.update` to generate this file.


ondrasej:
I think I tried this and git still ignored changes made to the file, but ack. Creating the file is documented in README, so we don't really need it.

PR #4, opened 2023-05-28
Add documentation on obtaining training data

boomanaiden154:
This patch adds in documentation on obtaining training data from the BHive dataset, makes some notes about other datasets, and marks out future directions (particularly extraction from ELF binaries). This spells out a couple things in the process that aren't too difficult to figure out but might not be very obvious (such as running all of the tooling through bazel), especially for those unfamiliar with the tooling/build system.

### DIFF
--- a/g3doc/obtaining-training-data.md
+++ b/g3doc/obtaining-training-data.md
@@ -1,0 +1,58 @@
+# Obtaining Training Data
+
+This document gives an overview of how to use the Gematria tooling to
+parse/process training data from multiple sources.
+
+## Parsing training data from the BHive dataset
+
+The publically available [BHive](https://github.com/ithemal/bhive) dataset
+has about 300k basic blocks along with throughput data for Ivy Bridge, Haswell,
+and Skylake in a CSV format. Gematria provides native tooling to process these
+CSV files into `.tfrecord` files that can be used with Gematria models.
+
+Every line in the BHive CSV files has the format
+`<machine code of BB in hex>,<throughput in cycles>`. This is referred to as
+the BHive format. The dataset contains one CSV file per mircoarchitecture.
+
+To process BHive throughput data for the Intel Skylake microarchitecture:
+
+1. Download the corresponding BHive CSV:
+
+```bash
+curl -L https://raw.githubusercontent.com/ithemal/bhive/5f1d50077ac0779fd227b261dcf517862c7104bd/benchmark/throughput/skl.csv > skl.csv
+```
+
+2. Convert the downloaded CSV file into a `.tfrecord` file:
+
+```bash
+bazel run //gematria/datasets/python:import_from_bhive -- \
+    --gematria_input_csv=skl.csv \
+    --gematria_output_tfrecord=skl.tfrecord \
+    --gematria_throughput_source_name="bhive: skl"
+```
+
+This will create a `skl.tfrecord` file containing imported basic blocks that
+can be used for further downstream analysis. The `--gematria_input_csv` and
+`--gematria_output_tfrecord` flags specify the path to the input CSV and the
+path to the output `.tfrecord` files respectively. The
+`--gematria_throughput_source_name` flag should be used to specify the method
+of data collection. The protobuf data format that Gematria uses allows for
+storing throughput information for multiple different microarchitectures for
+the same basic block within a single protobuf. This allows for having a single
+dataset with throughput data for all known microarchitectures that handles
+missing data from specific sources gracefully. Typically labels like
+`bhive: skl`, `bhive: hsw`, and `bhive: ivb` will be used in the same dataset
+(one per microarchitecture) and then a model will be trained with multi-task
+training.
+
+## Other CSV-based formats
+
+There are other projects that also contain basic blocks and their associated
+throughput data such Andreas Abel's [uica-eval](https://github.com/andreas-abel/uiCA-eval).
+However, these projects might use slightly different CSV formats and additional
+preprocessing may be needed to import them.
+
+## Parsing training data from ELF executables
+
+TODO(boomanaiden154): Write documentation on this section once tooling for
+extracting BBs from ELF binaries lands.


g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
+# Obtaining Training Data
+
+This document gives an overview of how to use the Gematria tooling to
+parse/process training data from multiple sources.
+
+## Parsing training data from the BHive dataset
+
+The publically available [BHive](https://github.com/ithemal/bhive) dataset
+has a decent amount of training data available in CSV format. Gematria
```
ondrasej, line 9: I'd replace "a decent" with "~300k basic blocks and their throughput for multiple microarchitectures".

ondrasej, line 9: s/CSV format/a CSV format/

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+provides native tooling to process these CSV files into `.tfrecord` files
+that can be used for further processing.
```
ondrasej, line 11: ...that can be used with Gematria models.

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+
+For example, to process BHive throughput data for the Intel Skylake
+microarchitecture, we can begin by downloading one of the BHive CSVs:
+
+```bash
+curl -L https://raw.githubusercontent.com/ithemal/bhive/5f1d50077ac0779fd227b261dcf517862c7104bd/benchmark/throughput/skl.csv > skl.csv
+```
+
+If you look at the file, you'll notice that every line is formatted as
```
ondrasej, line 20: I'd remove everything before "Every line is ..."

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+parse/process training data from multiple sources.
+
+## Parsing training data from the BHive dataset
+
+The publically available [BHive](https://github.com/ithemal/bhive) dataset
+has a decent amount of training data available in CSV format. Gematria
+provides native tooling to process these CSV files into `.tfrecord` files
+that can be used for further processing.
+
+For example, to process BHive throughput data for the Intel Skylake
```
ondrasej, line 13: We should start by saying that the data set contains one CSV file per microarchitecture (and we can say that for example, the data for Skylake are in `skl.csv`.

That should make it easier to understand for the readers why we're picking a single microarchitecture for the example instead of putting everything to the same data set.

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+If you look at the file, you'll notice that every line is formatted as
+`<machine code of BB in hex>,<throughput in cycles>>`. This is referred to as
+the BHive format. There are other projects composing similar datasets (such
+as Adreas Abel's [uica-eval](https://github.com/andreas-abel/uiCA-eval) which
+contains several improvements over the original BHive dataset) that use similar
+but not necessarily the same CSV format. Note that Gematria only supports the
+BHive format outlined above, so some preprocessing might be required depending
+upon the input data format.
```
ondrasej, line 27: I'd put this to a subsection at the end of the BHive section. Something like

```
## Other CSV-based formats

Other projects, such as Andreas Abel's [uica-eval](...) use a similar format. However, there are subtle differences between the formats, and additional pre-processing may be needed.
```

I didn't look at the data format in `uica-eval`, but if it is sufficiently close, adding an import tool might be less work than explaining why we don't support it directly 😄  I can check it out when I'm back from vacations.

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+
+To process this into a `.tfrecord` that can be used, the `import_from_bhive`
+script within Gematria can be run through Bazel:
+
+```bash
+bazel run //gematria/datasets/python:import_from_bhive -- \
+    --gematria_input_csv=skl.csv \
+    --gematria_output_tfrecord=skl.tfrecord \
+    --gematria_throughput_source_name="bhive: skl"
+```
```
ondrasej, line 37: I'd prefer to put both command-lines next to each other, so that the user can follow them like instructions, something like

1. Download the CSV file \
   ```bash
   curl -L https://raw.githubusercontent.com/ithemal/bhive/5f1d50077ac0779fd227b261dcf517862c7104bd/benchmark/throughput/skl.csv > skl.csv
   ```
3. Convert the CSV to `.tfrecord` \
   ```bash
   bazel run //gematria/datasets/python:import_from_bhive -- \
     --gematria_input_csv=skl.csv \
     --gematria_output_tfrecord=skl.tfrecord \
     --gematria_throughput_source_name="bhive: skl"
   ```

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+
+This will create a `skl.tfrecord` file containing imported basic blocks that
+can be used for further downstream analysis. The `--gematria_input_csv` and
+`--gematria_output_tfrecord` flags specify the path to the input CSV and the
+path to the output `.tfrecord` files respectively. The
+`--gematria_throughput_source_name` flag should be used to specify the method
+of data collection. More information on this flag is available in the protobuf
+definition in `/gematria/proto/throughput.proto`.
```
ondrasej, line 45: No action needed here: yeah, I need to add a tool to merge .tfrecord files from different micro-architectures.

ondrasej:
Thanks for writing this!

I've added a bunch of comments on style and organisation (and I might have more once it they are addressed), but overall I think the doc has what it needs to have and I think it will be very useful for new users.

ondrasej (changes requested):

g3doc/obtaining-training-data.md
boomanaiden154: Makes sense. Once we have that in place, I think everything should be in the repository to fully train a multi-microarchitecture GRANITE model. I'll try and write some documentation on the end-to-end pipeline tutorial style as I worked through the steps when I get a chance.

g3doc/obtaining-training-data.md
boomanaiden154: It's pretty close. The CSV format in `uica-eval` just has a couple extra columns. The main thing that I think would be needed would be flags to specify the column indices for the BB hex and throughput value. I should be able to get a patch up soon for that.

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+The publically available [BHive](https://github.com/ithemal/bhive) dataset
+has about 300k basic blocks along with throughput data for Ivy Bridge, Haswell,
+and Skylake in a CSV format. Gematria provides native tooling to process these
+CSV files into `.tfrecord` files that can be used with Gematria models.
+
+Every line in the BHive CSV files has the format
+`<machine code of BB in hex>,<throughput in cycles>>`. This is referred to as
+the BHive format. The dataset contains one CSV file per mircoarchitecture.
+To process BHive throughput data for the Intel Skylake
+microarchitecture, we can begin by downloading one of the BHive CSVs:
```
ondrasej, line 17: I'd even do this as
To process BHive throughput data for the Intel Skylake microarchitecture:

1. download the corresponding BHive CSV:
   ```
   ...
   ```
2. Convert the downloaded CSV file into a `.tfrecord` file:
   ```
   ...
   ```

g3doc/obtaining-training-data.md
```diff
@@ -0,0 +1,50 @@
...
+    --gematria_output_tfrecord=skl.tfrecord \
+    --gematria_throughput_source_name="bhive: skl"
+```
+
+This will create a `skl.tfrecord` file containing imported basic blocks that
+can be used for further downstream analysis. The `--gematria_input_csv` and
+`--gematria_output_tfrecord` flags specify the path to the input CSV and the
+path to the output `.tfrecord` files respectively. The
+`--gematria_throughput_source_name` flag should be used to specify the method
+of data collection. More information on this flag is available in the protobuf
```
ondrasej, line 37: The throughput source name is there to disambiguate throughput information for different microarchitectures.

The proto data format allows storing throughput information from multiple sources (measurement methods, microarchitectures, ...) for the same basic block in the same proto. This way, we can have a single data set with throughput data for all known microarchitectures. The format (and the code around it) is flexible enough that it works even when some blocks in the data set have data only from some sources.

Typically, we'd have `bhive: skl`, `bhive: hsw`, `bhive: ivb`, ... in the same data set and train it using multi-task training.

ondrasej (changes requested):

g3doc/obtaining-training-data.md
boomanaiden154: I added a note on this info. I was trying not to duplicate information between the protobuf definition and the documentation, but just pointing to the code from the docs is probably not a super great strategy. Thanks for the detailed explanation!

ondrasej (approved):

PR #6, opened 2023-06-02
Add requirements.txt and update documentation

boomanaiden154:
This patch adds in a full requirements.txt representing the transitive closure for the dependencies as this seems to be the proper way to utilize the Bazel compile_pip_requirements command based on looking at other projects. This patch also updates the appropriate documentation pertaining to this.

I looked at the following repositories to see what others were doing:
* https://github.com/nanopb/nanopb/tree/master
* https://github.com/google-research/nisaba/tree/main
* https://github.com/nanopb/nanopb

Something should probably be added into the Bazel py_rules repository noting that this is the way it should be used.

Should probably also get some precommit CI together at some point using Github actions to verify that my changes don't end up breaking anything.


ondrasej:
I remember trying with the full `requirements.txt` checked in at first, but got some compatibility issues when I was switching from one machine to other (and perhaps from Linux/x86 to OS X/ARM).

Let's check that this works on both before submitting.

boomanaiden154:
That would make sense. Bazel does generate the `requirements.txt` with `system_platform` annotations, but it doesn't add in any of the MacOS specific packages, just adding the annotations about what should be on Linux, platform restrictions (i.e., `tensorflow-macos` isn't on there). Generating the file on Mac and then manually copying over the hashes should probably work, but I'd prefer to avoid that solution. I'm thinking another `requirements.txt` for MacOS specifically should work?

ondrasej:
Maybe compiling it with `pip-compile` does a better job than bazel? I didn't try, just thinking of possible solutions.

boomanaiden154:
Seems like this has been superseded by 55c2440b515fff0d710736a462e080b4ca92607d. That doesn't give us a transitive closure which make some dependencies not reproducible, but I guess if that becomes an issue we can look into solutions (maybe something like `pip-env`).

ondrasej:
I've discussed this with https://github.com/Mizux (the maintainer of @google/or-tools), and his experience was that the hashes in requirements.txt depend on the version of Python used, target platform, and a number of other factors, and any fixed set of hashes is broken for _some_ platform.

Let's see how keeping just `requirements.in` works for us; if it breaks things more often than not, we can go back to having instructions on how to generate `requirements.txt` (using PIP or other standard tools), or submit one with specific versions, but without hashes.

boomanaiden154:
That makes sense. We've definitely run into some trouble getting that to work with `ml-compiler-opt` across the three python versions that we support.

The plan to only switch back if we run into issues sounds perfect.

PR #7, opened 2023-06-03
Add support for importing CSVs with differing formats

boomanaiden154:
Other datasets such as Andreas Abel's uica-eval use slightly different CSV formats and there's no real reason not to support them since the only thing that differs is the column layout. This patch adds in two new flags that allow specifying the columns of the basic block hex and throughput data.

Should be mostly complete, but is in desperate need of formatting (which is why I've marked it as a draft).

### DIFF
--- a/gematria/datasets/bhive_importer.cc
+++ b/gematria/datasets/bhive_importer.cc
@@ -14,6 +14,7 @@
 
 #include "gematria/datasets/bhive_importer.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <memory>
 #include <string_view>
@@ -25,7 +26,7 @@
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/numbers.h"
-#include "absl/strings/str_cat.h"
+#include "absl/strings/str_format.h"
 #include "absl/strings/str_split.h"
 #include "absl/types/span.h"
 #include "gematria/basic_block/basic_block_protos.h"
@@ -95,16 +96,26 @@ BHiveImporter::BasicBlockProtoFromMachineCodeHex(
 
 absl::StatusOr<BasicBlockWithThroughputProto> BHiveImporter::ParseBHiveCsvLine(
     std::string_view source_name, std::string_view line,
+    size_t machine_code_hex_column_index, size_t throughput_column_index,
     double throughput_scaling /*= 1.0*/, uint64_t base_address /*= 0*/) {
   const absl::InlinedVector<std::string_view, 2> columns =
       absl::StrSplit(line, ',');
-  if (columns.size() != 2) {
-    return absl::InvalidArgumentError(
-        absl::StrCat("Expected `line` to have 2 columns, found ",
-                     columns.size(), ": '", line, "'"));
+  const int min_required_num_columns =
+      std::max(machine_code_hex_column_index, throughput_column_index) + 1;
+  if (columns.size() < min_required_num_columns) {
+    return absl::InvalidArgumentError(absl::StrFormat(
+        "Expected `line` to have at least %d columns, found %d: %s",
+        min_required_num_columns, columns.size(), line));
   }
-  const std::string_view machine_code_hex = columns[0];
-  const std::string_view throughput_str = columns[1];
+  if (machine_code_hex_column_index == throughput_column_index) {
+    return absl::InvalidArgumentError(absl::StrFormat(
+        "Expected machine code column and throughput column indices to be "
+        "different for `line`, but were both %d: %s",
+        machine_code_hex_column_index, line));
+  }
+  const std::string_view machine_code_hex =
+      columns[machine_code_hex_column_index];
+  const std::string_view throughput_str = columns[throughput_column_index];
 
   BasicBlockWithThroughputProto proto;
   absl::StatusOr<BasicBlockProto> block_proto_or_status =
--- a/gematria/datasets/bhive_importer.cc
+++ b/gematria/datasets/bhive_importer.cc
@@ -110,7 +110,7 @@ absl::StatusOr<BasicBlockWithThroughputProto> BHiveImporter::ParseBHiveCsvLine(
   if (machine_code_hex_column_index == throughput_column_index) {
     return absl::InvalidArgumentError(absl::StrFormat(
         "Expected machine code column and throughput column indices to be "
-        "different for `line`, but were both %d: %s",
+        "different, but were both %d: %s",
         machine_code_hex_column_index, line));
   }
   const std::string_view machine_code_hex =
--- a/gematria/datasets/bhive_importer.h
+++ b/gematria/datasets/bhive_importer.h
@@ -70,6 +70,7 @@ class BHiveImporter {
   // `base_address` as the address of the first instruction in the basic block.
   absl::StatusOr<BasicBlockWithThroughputProto> ParseBHiveCsvLine(
       std::string_view source_name, std::string_view line,
+      size_t machine_code_hex_column_index, size_t throughput_column_index,
       double throughput_scaling = 1.0, uint64_t base_address = 0);
 
  private:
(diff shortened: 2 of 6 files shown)


boomanaiden154:
Just reformatted everything using the new tooling that has been set up. Should be ready to start review on this patch now.

gematria/datasets/bhive_importer.cc
```diff
@@ -95,16 +95,20 @@ BHiveImporter::BasicBlockProtoFromMachineCodeHex(
...
     std::string_view source_name, std::string_view line,
-    double throughput_scaling /*= 1.0*/, uint64_t base_address /*= 0*/) {
+    size_t machine_code_hex_column_index /*= 0*/,
+    size_t throughput_column_index /*= 1*/, double throughput_scaling /*= 1.0*/,
+    uint64_t base_address /*= 0*/) {
   const absl::InlinedVector<std::string_view, 2> columns =
       absl::StrSplit(line, ',');
-  if (columns.size() != 2) {
+  if (columns.size() <= machine_code_hex_column_index ||
+      columns.size() <= throughput_column_index) {
```
ondrasej, line 104: I'd use something like
```c++
const int min_required_num_columns = std::max(
    machine_code_hex_column_index, throughput_column_index) + 1;
if (columns.size() < min_required_num_columns) {
  return absl::InvalidArgumentError(
    absl::StrFormat("Expected `line` to have at least %d columns, found %d: %s",
                    min_required_num_columns, columns.size(), line));
}
```

gematria/datasets/bhive_importer.h
```diff
@@ -70,7 +70,9 @@ class BHiveImporter {
   // `base_address` as the address of the first instruction in the basic block.
   absl::StatusOr<BasicBlockWithThroughputProto> ParseBHiveCsvLine(
       std::string_view source_name, std::string_view line,
-      double throughput_scaling = 1.0, uint64_t base_address = 0);
+      size_t machine_code_hex_column_index = 0,
+      size_t throughput_column_index = 1, double throughput_scaling = 1.0,
```
ondrasej, line 74: If we give the column indices almost always, I'd remove the defaults.

Maybe we should remove defaults also from the scaling and the base address, default values just complicate things, but that's out of scope of this PR.

gematria/datasets/bhive_importer_test.cc
```diff
@@ -182,27 +182,37 @@ TEST_F(BHiveImporterTest, MultipleInstructions) {
...
-          /*throughput_scaling=*/kScaling, /*base_address=*/100),
-      IsOkAndHolds(EqualsProto(kExpectedBasicBlockProto)));
+  EXPECT_THAT(x86_bhive_importer_->ParseBHiveCsvLine(
+                  kSourceName,
+                  "4829d38b44246c8b54246848c1fb034829d04839c3,207.000000", 0, 1,
+                  /*throughput_scaling=*/kScaling, /*base_address=*/100),
+              IsOkAndHolds(EqualsProto(kExpectedBasicBlockProto)));
 }
 
 TEST_F(BHiveImporterTest, NoCommaOnLine) {
```
ondrasej, line 192: Let's rename this one to InvalidColumnIndices (and add another failing case with more columns).

gematria/datasets/python/import_from_bhive.py
```diff
@@ -66,6 +66,16 @@
     'x86_64',
     'The LLVM triple used for disassembling the instructions in the data set.',
 )
+_MACHINE_CODE_HEX_COLUMN_INDEX = flags.DEFINE_integer(
+    'machine_code_hex_column_index',
+    '0',
+    'The index to the the machine code hex column in the input CSV file.',
```
ondrasej, line 72: Nit: s/to/of/

Same applies to the following flag.

ondrasej (changes requested):
The PR looks good overall. I've added a few comments & I'll approve once they are resolved.

gematria/datasets/bhive_importer.h
boomanaiden154: I've removed the default column indices in this patch. I'll probably write a follow up removing the throughput scaling and base addresses.

gematria/datasets/bhive_importer.cc
```diff
@@ -95,16 +95,20 @@ BHiveImporter::BasicBlockProtoFromMachineCodeHex(
 
 absl::StatusOr<BasicBlockWithThroughputProto> BHiveImporter::ParseBHiveCsvLine(
     std::string_view source_name, std::string_view line,
-    double throughput_scaling /*= 1.0*/, uint64_t base_address /*= 0*/) {
+    size_t machine_code_hex_column_index /*= 0*/,
+    size_t throughput_column_index /*= 1*/, double throughput_scaling /*= 1.0*/,
+    uint64_t base_address /*= 0*/) {
```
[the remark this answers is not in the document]
boomanaiden154: I added a check in for this.

gematria/datasets/python/bhive_importer.cc
```diff
@@ -102,7 +102,9 @@ PYBIND11_MODULE(bhive_importer, m) {
       .def(  //
           "basic_block_with_throughput_proto_from_csv_line",
           &BHiveImporter::ParseBHiveCsvLine, py::arg("source_name"),
-          py::arg("line"), py::arg("throughput_scaling") = 1.0,
+          py::arg("line"), py::arg("machine_code_hex_column_index") = 0,
+          py::arg("throughput_column_index") = 1,
```
ondrasej, line 106: Let's also remove the defaults from here.

ondrasej (changes requested):
A few last comments, I'll approve & merge once it is resolved.

Btw. when doing the code reviews, please make the changes as normal commits in the branch rather than amend and force push. Adding new commits allows me to review the code incrementally (i.e. see what I already reviewed).

At the end, there is an option to clean up the commit history of the branch (e.g. squash all the commits into one, and update the commit message or use that from the PR), so there's no need to worry about making a mess in the commit history. And the incremental reviews can be a lot more efficient, especially for larger changes.

gematria/datasets/python/import_from_bhive.py
```diff
@@ -77,6 +77,14 @@
     'The index of the throughput value column in the input CSV file.',
 )
 
+flags.register_multi_flags_validator(
+    ['machine_code_hex_column_index', 'throughput_column_index'],
+    lambda flags_dict: flags_dict['machine_code_hex_column_index']
+    != flags_dict['throughput_column_index'],
+    'Expected machine code column and throughput column indices to be'
+    ' different',
+)
```
ondrasej, line 86: The [Google style guide](https://google.github.io/styleguide/pyguide.html#210-lambda-functions) doesn't like "complex" lambdas (and it's very inclusive about what is complex).

A very common pattern for the validators is to use a decorator and a normal function:
```
@flags.multi_flags_validator(
    ['machine_code_hex_column_index', 'throughput_column_index'],
    message='Expected machine code column and throughput column indices to be different',
)
def _validate_input_columns(flags_dict):
  return flags_dict['machine_code_hex_column_index'] != flags_dict['throughput_column_index']
```

ondrasej (changes requested):
One last comment (I think), then I'll approve.

gematria/datasets/python/import_from_bhive.py
```diff
@@ -77,13 +77,19 @@
     'The index of the throughput value column in the input CSV file.',
 )
 
-flags.register_multi_flags_validator(
+
+@flags.multi_flags_validator(
     ['machine_code_hex_column_index', 'throughput_column_index'],
```
ondrasej, line 82: There's a way to avoid using string literals for flag names: instead of `'machine_code_hex_column_index'`, it's better to use the flag holder to get the name with `_MACHINE_CODE_HEX_COLUMN_INDEX.name`.

Similar for other flag names.

ondrasej (changes requested):
There have been way too many "one last comment", and I'm really sorry about it. But I just realised there's one more thing to fix.

But this should be the really last one, and if there's more, I'll just approve and then do the fix 😄

ondrasej (approved):

PR #9, opened 2023-06-07
Add clang-format config and documentation

boomanaiden154:
No changed files since these seems to be the format that was used internally. However, a config file still needs to be added as the default style for `clang-format` isn't `Google`, and it's good to have some documentation on how to use the formatting tools.


ondrasej:
Cool, thanks for adding this!

PR #12, opened 2023-06-18
Add formatting jobs to CI

boomanaiden154:
As a public repository, Gematria gets unlimited free minutes for running jobs on Github CI. We've had good success using Github actions to ensure code is formatted properly/tests pass in MLGO, so I'm opening up this patch to do the same with Gematria so that it's less likely developers miss something.

Note that I'm using external dependencies here to make the setup easier, which may not be super desirable due to issues like maintenance concerns, but migrating off of them should be trivial. Something to note, however.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,36 @@
+name: Gematria CI
+
+on: [push, repository_dispatch, pull_request]
+
+jobs:
+  check-python-formatting:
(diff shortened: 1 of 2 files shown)


ondrasej:
This is really cool, thanks!

Let's see how it works in practice, but unless the formatters change significantly, this should not generate a lot of maintenance work.

PR #13, opened 2023-06-23
Update Dockerfile after dependency installation change

boomanaiden154:
Recently we moved from declaring an entire transitive closure in requirements.txt (and generating that with Bazel) to declaring just the top-level packages along with version information in requirements.in. This change brought with it a different set of commands for setup which breaks the existing Dockerfile. This patch fixes the Dockerfile to match the new setup.


ondrasej:
Thanks for fixing this!

PR #14, opened 2023-06-23
Experiment - access pattern benchmarks

virajbshah:
# Experiment - access pattern benchmarks

Adds experiments/access_pattern_bm.

This directory contains a set of microbenchmarks whose purpose 
is to serve as a sanity check of sorts for cache hit/miss performance 
related predictions from coming experiments/changes to the cost model.

### Build

The benchmarks can be built using
```bash
bazel build //gematria/experiments/access_pattern_bm:all
```
and built binaries for each benchmark can be found down the corresponding 
path in `bazel-bin`.

### Benchmarks

Currently, there are microbenchmarks for the following access patterns:

 * Pointer chasing (linked lists, graphs): `linked_list_bm`
 * Contiguous chunks of memory: `contiguous_matrix_bm`
 * Vector-of-vector type accesses: `vec_of_vec_matrix_bm`
 * Iterating over any STL-like container (multiset, list, deque, etc): `stl_container_bm`
 * Iterating over any STL-like associative container (map, unordered_map, etc): `stl_container_bm`


gematria/experiments/access_pattern_bm/stl_container.h
```diff
@@ -0,0 +1,42 @@
...
+static std::uniform_int_distribution<int> distribution(0, 1023);
+
+template <typename Container>
+Container *CreateRandomSTLContainer(std::size_t size) {
+  static_assert(std::is_same<typename Container::value_type, int>::value,
+                "Container must hold `int`s.");
+
+  auto elements = std::vector<int>(size);
+  for (int &element : elements) element = distribution(generator);
+  auto container = new Container(elements.begin(), elements.end());
```
ondrasej, line 21: Simpler:
```
Container container;
auto inserter = std::inserter(container, container.end());
for (int i = 0; i < size; ++i, ++inserter) {
  *inserter = distribution(generator);
}
```

This should work for pretty much all non-associative STL containers.

gematria/experiments/access_pattern_bm/contiguous_matrix.h
```diff
@@ -0,0 +1,14 @@
...
+
+namespace gematria {
+
+int *CreateRandomContiguousMatrix(const std::size_t size);
+void FlushContiguousMatrixFromCache(int *matrix, std::size_t size);
+void DeleteContiguousMatrix(int *matrix);
+
+}  // namespace gematria
+
+#endif
```
ondrasej, line 14: _Read the other comments before acting on this one, if we merge all into one file, no action would be needed here._

The `endif` should have a comment that says this is for the header guard. Here:

```
#endif  // GEMATRIA_EXPERIMENTS_ACCESS_PATTERN_BM_CONTIGUOUS_MATRIX_H_
```

Same for the other headers.

WORKSPACE
```diff
@@ -42,6 +42,12 @@ git_repository(
     tag = "release-1.12.1",
 )
 
+git_repository(
+    name = "com_google_benchmark",
```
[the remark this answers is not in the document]
virajbshah: I see :+1: . I picked up the `com_google_benchmark` from fleetbench's source code, along with the older `BENCHMARK_TEMPLATE` syntax.

ondrasej:
Once you push your updates, please set me as "Assignee" in the PR, so that I know I need to take a look.

gematria/experiments/access_pattern_bm/BUILD
```diff
@@ -0,0 +1,95 @@
...
+
+cc_test(
+    name = "linked_list_bm",
+    size = "small",
+    srcs = ["linked_list_bm.cc"],
+    deps = [
+        ":linked_list",
+        "@com_google_benchmark//:benchmark",
+    ],
+)
```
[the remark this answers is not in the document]
virajbshah: I don't quite expect the individual cc_libraries to be re-used elsewhere, but left them separate considering that maybe the need arises to have some other variation of the microbenchmark using the same code but handling things somewhat differently. Would it be better to have them be merged now, but split up later in case it becomes a requirement?

gematria/experiments/access_pattern_bm/contiguous_matrix_bm.cc
```diff
@@ -0,0 +1,85 @@
...
+BENCHMARK(BM_FlushContiguousMatrixFromCache)->Range(1 << 4, 1 << 12);
+
+static void BM_ContiguousMatrix_NoFlush(benchmark::State &state) {
+  const std::size_t size = state.range(0);
+
+  // Create a random matrix
+  auto matrix = CreateRandomContiguousMatrix(size);
+  // auto mock = CreateRandomContiguousMatrix(size);
+
+  int sum = 0;
```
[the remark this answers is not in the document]
virajbshah: As in, move the code that creates and destroys the container in question into the benchmark's `for (auto _ : state)` loop? I may be misunderstanding how `benchmark` works, but would that not "muddle up" the calculation by repeating those construction/destruction operations as part of the benchmark iteration process? If at all we also want to measure cache miss rates while creating containers, wouldn't that best be done as a separate microbenchmark?

gematria/experiments/access_pattern_bm/contiguous_matrix_bm.cc
```diff
@@ -0,0 +1,85 @@
...
+  for (auto _ : state) {
+    FlushContiguousMatrixFromCache(matrix, size);
+
+    // Loop over the matrix, doing some dummy
+    // operations along the way.
+    for (int i = 0; i < size; ++i) {
+      for (int j = 0; j < size; ++j) {
+        sum += matrix[size * i + j];
+      }
+    }
```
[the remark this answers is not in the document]
virajbshah: I ran both variants (this, and having both `i` and `j` go backwards) though a quick `perf stat` test, not expecting too much of a difference - but very interestingly, the forwards variant was measured to have 12,591,020 `mem_load_retired.l3_miss`es as compared to the backwards variant's 6,749,696 over the same 4s. Even accounting for the fact that each microbenchmark was not executed for the same number of iterations owing to how `benchmark` handles that, I wouldn't expect both measurements to vary by a factor that is  so strangely close to 2.

I think it would be really fun to understand what microarchitectural details result in that, and how.

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -0,0 +1,37 @@
...
+  const char *ptr = (const char *)matrix;
+  const char *end = (const char *)(matrix + (size + 1) * sizeof(int));
+
+  _mm_mfence();
+  while (ptr <= end) {
+    _mm_clflushopt(ptr);
+    ptr += line_size;
+  }
+  _mm_mfence();
+}
```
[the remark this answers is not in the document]
virajbshah: This makes sense to me, but I am not sure if it'll clean up the code in its current state given that we are only flushing multiple container elements simultaneously in `contiguous_matrix_bm` since, as far as I am aware, that is the only place where elements are guaranteed to be stored contiguously.

We can also isolate the `-mclflushopt` dependency by writing a function that flushes one container element (given that it fits on a single cache line), but that would end up being a simple wrapper around `_mm_clflushopt`, which I found to be slightly clunky. Would doing so be a better idea?

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  return matrix;
+}
+
+void FlushContiguousMatrixFromCache(std::unique_ptr<int[]> &matrix,
+                                    const std::size_t size) {
+  constexpr int line_size = ABSL_CACHELINE_SIZE;
+  const char *ptr = (const char *)matrix.get();
+  const char *end = (const char *)(matrix.get() + (size + 1) * sizeof(int));
```
ondrasej, line 45: Careful, `matrix.get() + (size + 1)` will do pointer arithmetics in int size (it's done before the conversion to `const char*`), so end will end up at address `ptr + (size + 1) * sizeof(int) * sizeof(int)`.

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  return matrix;
+}
+
+void FlushContiguousMatrixFromCache(std::unique_ptr<int[]> &matrix,
+                                    const std::size_t size) {
+  constexpr int line_size = ABSL_CACHELINE_SIZE;
+  const char *ptr = (const char *)matrix.get();
```
ondrasej, line 44: Very nit: I'd call this `begin` or `start` to match with the `end` below (generally, when there is a variable called `end`, the opposite should be called `begin`).

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -0,0 +1,55 @@
...
+  const char *end = (const char *)(matrix.get() + (size + 1) * sizeof(int));
+
+  _mm_mfence();
+  while (ptr <= end) {
```
ondrasej, line 48: Optional, but recommended
```
for (; ptr <= end; ptr += line_size) {
  // ...
}
```

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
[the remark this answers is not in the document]
ondrasej: Ack. Let's keep it the way it is.

That said, it's probably better to use `-mclflushopt` on all targets in this directlry, because it might affect which defined are present in the preprocessor, and as a consequence, make some includes (namely `<immintr.h>` look different when they are parsed for different bazel targets).

gematria/experiments/access_pattern_bm/contiguous_matrix.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef GEMATRIA_EXPERIMENTS_ACCESS_PATTERN_BM_CONTIGUOUS_MATRIX_H_
+#define GEMATRIA_EXPERIMENTS_ACCESS_PATTERN_BM_CONTIGUOUS_MATRIX_H_
+
+#include <memory>
+
+namespace gematria {
+
+std::unique_ptr<int[]> CreateRandomContiguousMatrix(std::size_t size);
+void FlushContiguousMatrixFromCache(std::unique_ptr<int[]> &matrix,
+                                    std::size_t size);
```
ondrasej, line 24: Please add a comment whether `size` is in the number of ints or in bytes.

gematria/experiments/access_pattern_bm/contiguous_matrix.h
```diff
@@ -0,0 +1,28 @@
...
+
+}  // namespace gematria
+
+#endif
```
ondrasej, line 28: Nit: Add a comment `// GEMATRIA_EXPERIMENTS_ACCESS_PATTERN_BM_CONTIGUOUS_MATRIX_H_` to mark this as the end of the header guard.

gematria/experiments/access_pattern_bm/contiguous_matrix_test.cc
```diff
@@ -0,0 +1,91 @@
...
+}
+
+BENCHMARK(BM_FlushContiguousMatrixFromCache)->Range(1 << 4, 1 << 12);
+
+void BM_ContiguousMatrix_NoFlush(benchmark::State &state) {
+  const std::size_t size = state.range(0);
+
+  // Create a random matrix.
+  auto matrix = CreateRandomContiguousMatrix(size);
+#ifdef BALANCE_FLUSHING_TIME
```
ondrasej, line 42: Nit+optional: I'd recommend changing the use of `BALANCE_FLUSHING_TIME` to

Somewhere in some header:
```
inline constexpr bool kBalanceFlushingTime =
#ifdef BALANCE_FLUSHING_TIME
    true;
#else
    false;
#endif  // BALANCE_FLUSHING_TIME
```
and then in code use
```
std::unique_ptr<int[]> mock;
if (kBalanceFlushingTime) {
  mock = CreateRandomCongiguousMatrix(size);
}
// ...
if (kBalanceFlushingTime) {
  FLushContiguousMatrixFromCache(mock, size);
}
```

The advantage is that most of the code will be parsed by the compiler in any compile and you'd get warned about errors even in the de-activated parts. But at the same time, since the value of `kBalanceFlushingTime` is known at compile time (it's a `constexpr`), the compiler will be able to remove the disabled part via dead code elimination.

gematria/experiments/access_pattern_bm/contiguous_matrix_test.cc
```diff
@@ -0,0 +1,91 @@
...
+  auto mock = CreateRandomContiguousMatrix(size);
+#endif
+
+  int sum = 0;
```
ondrasej, line 46: I'd move this inside the loop. With `DoNotOptimize()`, we don't need to care where it is defined (and we wouldn't need the line `sum = 0;` at the end of the loop).

gematria/experiments/access_pattern_bm/contiguous_matrix_test.cc
```diff
@@ -0,0 +1,91 @@
...
+
+    benchmark::DoNotOptimize(sum);
+    sum = 0;
+  }
+}
+
+BENCHMARK(BM_ContiguousMatrix_NoFlush)->Range(1 << 4, 1 << 12);
+
+void BM_ContiguousMatrix_Flush(benchmark::State &state) {
+  const std::size_t size = state.range(0);
```
ondrasej, line 67: All comments from above also apply to this function. And to other functions below.

gematria/experiments/access_pattern_bm/linked_list.cc
```diff
@@ -0,0 +1,62 @@
...
+    current = current->next;
+  }
+
+  auto deleter = [](Node *ptr) {
+    while (ptr) {
+      Node *temp = ptr->next;
+      delete ptr;
+      ptr = temp;
+    }
+  };
```
ondrasej, line 45: Optional: You could define this as a functor
```
struct NodeDeleter {
  void operator()(Node* node) {
    while (ptr) {
      Node *temp = ptr->next;
      delete ptr;
      ptr = temp;
    }
  }
};
```
and then use it without having to instantiate the deleter explicitly, e.g. as
```
return std::unique_ptr<Node, NodeDeleter>(head);
```

It has the additional advantage that NodeDeleter is empty, so the unique_ptr<> doesn't need to carry a function pointer with it (i.e. `sizeof(std::unique_ptr<Node, NodeDeleter>) == 8` while (`sizeof(std::unique_ptr<Node, void(*)(Node*)>) == 16`. It doesn't really matter here though.

gematria/experiments/access_pattern_bm/stl_assoc_container.cc
ondrasej, line 1: I think this file can be safely removed. Bazel should be fine with header-only libraries.

gematria/experiments/access_pattern_bm/stl_container.cc
ondrasej, line 1: Same here.

ondrasej (changes requested):
I've read through roughly the first half of the files and added my comments. I'll try to go through the rest tomorrow.

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
virajbshah: Right, that line will need some fixing. I was trying to make sure the array is flushed entirely even if its first element isn't aligned to a cache line for some reason and things like that, but made a much more serious goof-up even for the simple case.

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -0,0 +1,55 @@
...
+  for (int i = 0; i < size; ++i) {
+    for (int j = 0; j < size; ++j) {
+      matrix[size * i + j] = distribution(generator);
+    }
+  }
+
+  return matrix;
+}
+
+void FlushContiguousMatrixFromCache(std::unique_ptr<int[]> &matrix,
```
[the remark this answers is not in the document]
virajbshah: So, if I understand correctly, I should also similarly have all other `Flush...FromCache` functions take `const T*` for whatever `T` works for that particular benchmark since there is no ownership transfer going on there either, right?

virajbshah:
I think I've covered all changes requested so far. I don't quite like the way I've used macros for balancing the flushing time, seeing that style guides consider them as a bit of a no-no. Would it be better to pass it as an option to the benchmark binaries while running them from the command line? From what I've seen, doing this is somewhat "hacky" since the `main` function is set-up by `BENCHMARK_MAIN`.

Thanks for the review.

gematria/experiments/access_pattern_bm/BUILD
```diff
@@ -59,60 +70,45 @@ cc_library(
...
-    srcs = ["stl_container_test.cc"],
-    copts = select({
-        ":balance_flushing_time": ["-DBALANCE_FLUSHING_TIME"],
-        "//conditions:default": [],
-    }) + FEATURE_OPTS,  # Since templated functions using _mm_clflushopt are defined in the header
+    srcs = COMMON_TEST_HDRS + [
+        "stl_container.h",
+        "stl_container_test.cc",
+    ],
+    copts = BALANCE_FLUSHING_TIME_OPTS + FEATURE_OPTS,  # Since templated functions using _mm_clflushopt are defined in the header
```
ondrasej, line 97: Very nitpicky: Please add a period at the end of the comment (all comments that look like a sentence should use punctuation).

Same below.

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -38,16 +39,18 @@ std::unique_ptr<int[]> CreateRandomContiguousMatrix(const std::size_t size) {
...
+// Flushes matrices created by `CreateRandomContiguousMatrix`. Takes the matrix
+// itself, along with its size (given that it is a size x size square matrix,
+// holding size * size ints).
+void FlushContiguousMatrixFromCache(const void *matrix,
                                     const std::size_t size) {
   constexpr int line_size = ABSL_CACHELINE_SIZE;
-  const char *ptr = (const char *)matrix.get();
-  const char *end = (const char *)(matrix.get() + (size + 1) * sizeof(int));
+  const char *begin = reinterpret_cast<const char *>(matrix);
+  const char *end = begin + size * size * sizeof(int) / sizeof(char);
```
ondrasej, line 49: `sizeof(char)` is always one by definition. Which might sound a bit weird, but that's what the spec says: https://en.cppreference.com/w/cpp/language/sizeof

gematria/experiments/access_pattern_bm/contiguous_matrix.cc
```diff
@@ -38,16 +39,18 @@ std::unique_ptr<int[]> CreateRandomContiguousMatrix(const std::size_t size) {
...
 
   _mm_mfence();
-  while (ptr <= end) {
-    _mm_clflushopt(ptr);
-    ptr += line_size;
+  for (; begin <= end; begin += line_size) {
```
ondrasej, line 52: I'd go with `begin < end` here. `end` is the first byte that is _not_ in the matrix.

ondrasej (changes requested):
I've added a few more comments, but that should be the last ones.

Please ping me once they are resolved, and I'll approve (and merge).

ondrasej (approved):

PR #15, opened 2023-07-27
Add libpfm4 git repository to enable use of perf counters by benchmarks.

virajbshah:
Adds `libpfm4` and `foreign_rules_cc` (which is required to build `libpfm4`, which is configured to be built by GNU Make) to `WORKSPACE`. `libpfm4` is required by Google Benchmark when `//gematria/experiments/access_pattern_bm/...` targets are built with performance counters enabled.


WORKSPACE
```diff
@@ -48,6 +48,45 @@ git_repository(
...
+    name = "libpfm",
+    lib_source = ":pfm_srcs",
+    lib_name = "libpfm",
+    # this disables debug mode. In particular, this makes sure certain
+    # warning-as-error of libpfm aren't surfaced, which would break the build.
+    args = ["DBG="],
+    visibility = [
+        "//visibility:public",
+    ],
+)
```
ondrasej, line 84: I'd prefer to put this into a separate file (libpfm4.BUILD) that can be referenced from here , similar to `graph_nets.BUILD`, so that we have a bit of a separation between the list of repositories, and the build files for them.

WORKSPACE
```diff
@@ -48,6 +48,45 @@ git_repository(
...
+filegroup(
+    name = "pfm_srcs",
+    srcs = glob(["**"]),
+)
+
+make(
+    name = "libpfm",
+    lib_source = ":pfm_srcs",
+    lib_name = "libpfm",
+    # this disables debug mode. In particular, this makes sure certain
```
ondrasej, line 78: Nit: # This...

ondrasej (changes requested):
Overall, this looks good. I just have two minor comments, then I'll approve/merge.

boomanaiden154:
Is there any possibility that we would be able to just use the LLM Bazel third party rules to build libpfm? Adding it that way is a lot simpler:

```bazel
git_repository(
    name = "pfm",
    build_file = "@llvm-raw//utils/bazel/third_party_build:pfm.BUILD",
    remote = "https://git.code.sf.net/p/perfmon2/libpfm4",
    tag = "v4.12.1",
)
```

and makes it easy to work with parts of LLVM that need libpfm (which I'm planning on working on soon). The only major thing I can think of is that Google benchmark uses the name `libpfm` for the library whereas llvm just calls it `pfm`. Maybe there's a way to alias them (although `alias` seems to be specific to targets)?

ondrasej:
> Is there any possibility that we would be able to just use the LLM Bazel third party rules to build libpfm? Adding it that way is a lot simpler:
> 
> ```python
> git_repository(
>     name = "pfm",
>     build_file = "@llvm-raw//utils/bazel/third_party_build:pfm.BUILD",
>     remote = "https://git.code.sf.net/p/perfmon2/libpfm4",
>     tag = "v4.12.1",
> )
> ```
> 
> and makes it easy to work with parts of LLVM that need libpfm (which I'm planning on working on soon). The only major thing I can think of is that Google benchmark uses the name `libpfm` for the library whereas llvm just calls it `pfm`. Maybe there's a way to alias them (although `alias` seems to be specific to targets)?

+1, this sounds like the easiest way to get it working.

We already pull LLVM to the Gematria repo, so it's just a matter of putting all the reference in place.

virajbshah:
I ended up using the LLVM third party build rules for `libpfm4`, and changing Benchmark's build file to match LLVM's naming for the `libpfm4` target (`@libpfm//:libpfm` -> `@pfm//:pfm`) since it seemed to be the most painless way to do it.

ondrasej (changes requested):
Can we use a .patch file to change the BUILD from benchmark rather than keep a full copy?

WORKSPACE
```diff
@@ -42,10 +42,34 @@ git_repository(
...
+git_repository(
+    name = "rules_foreign_cc",
+    commit = "816905a078773405803e86635def78b61d2f782d",
+    remote = "https://github.com/bazelbuild/rules_foreign_cc.git",
+)
+
+load("@rules_foreign_cc//foreign_cc:repositories.bzl", "rules_foreign_cc_dependencies")
+
+rules_foreign_cc_dependencies()
+
```
ondrasej, line 66: I think we don't need this anymore.

ondrasej (changes requested):
I've added one more comment. I'll approve once it is resolved.

WORKSPACE
virajbshah: The build fails without it - I had actually mistakenly removed this and `rules_foreign_cc` in the commit before this one, thinking it was no longer needed since building `libpfm4` had been handed off to the LLVM third party rules, but it was only actually working because I hadn't `bazel clean --expunge`d before making sure it would build.

ondrasej: Ack. In that case, please add a comment that it is needed to build libpfm.

ondrasej (approved):
Feel free to merge once you've added the comment.

ondrasej (approved):

PR #16, opened 2023-08-08
Update documentation to correct config options for benchmarks

boomanaiden154:
Currently the documentation shows setting --pfm=true to get access to performance counters in the access pattern benchmarks, but Google benchmark only recognizes --pfm=1. This patch corrects the documentation for this.


boomanaiden154:
I looked into just making `--pfm=1` the default but couldn't get that to work after a lot of experimentation (but got fairly close). It should be fairly trivial to do with a patch to the benchmark bazel build if we go down that route.

ondrasej (approved):

PR #17, opened 2023-08-21
Improve access pattern benchmarks. 

virajbshah:
This patch
 * fixes a bug in `BM_STLAssocContainer_NoFlush`,
 * changes the flag for `balance_flushing_time` from `true` to `1` to match `pfm`,
 * adds calls to `PauseTiming` and `ResumeTiming` to exclude flushing overhead from all measurements,
 * and moves the version of Google Benchmark this depends on up to include the bug fix for use of `PauseTiming` and `ResumeTiming` with perf counters.


gematria/experiments/access_pattern_bm/stl_assoc_container_test.cc
```diff
@@ -81,9 +82,11 @@ void BM_STLAssocContainer_Flush(benchmark::State &state) {
   // Create a random associative container.
   auto container = CreateRandomSTLAssocContainer<Container>(size);
 
-  int sum = 0;
   for (auto _ : state) {
+    int sum = 0;
```
ondrasej, line 86: With this change, we don't need the `sum = 0;` on line 98 anymore.

(and for some reason GitHub doesn't let me comment directly on line 98).

gematria/experiments/access_pattern_bm/stl_container_test.cc
```diff
@@ -82,9 +84,11 @@ void BM_STLContainer_Flush(benchmark::State &state) {
   // Create a random container.
   auto container = CreateRandomSTLContainer<Container>(size);
 
-  int sum = 0;
   for (auto _ : state) {
+    int sum = 0;
```
ondrasej, line 88: Same comment applies here.

ondrasej (changes requested):
Just a few minor comments.

gematria/experiments/access_pattern_bm/stl_assoc_container_test.cc
ondrasej, line 86: Also remove line 98.

gematria/experiments/access_pattern_bm/stl_container_test.cc
ondrasej, line 88: Also remove line 100.

ondrasej (changes requested):

gematria/experiments/access_pattern_bm/stl_container_test.cc
virajbshah: Something strange seems to be going on there. I had removed the extra `sum = 0` in the previous commit itself, and I don't see that line on [my fork](https://github.com/virajbshah/gematria/blob/bm-measurement-changes/gematria/experiments/access_pattern_bm/stl_container_test.cc). The line is looks like it is still there in the PR diffs, but is shown as removed when I [compare](https://github.com/google/gematria/compare/main...virajbshah:gematria:bm-measurement-changes) google:main with my branch on my fork.

ondrasej: Weird. But it looks good here.

ondrasej (approved):

PR #19, opened 2023-10-13
Add a cmake build step in the CI

mtrofin:
The step heavily relies on ccache to dramatically reduce tflite and llvm build times. After the initial ccache warm-up, which would likely take 1hr or more, expect ~5 min or so build times.


ondrasej (approved):

boomanaiden154 (approved):
LGTM. There are a couple things that we'll probably want to add in the future, particularly running the tests and then the bazel build (which I can hack on finishing up).

This isn't entirely future proof/easily reproducible locally, but that shouldn't be an issue for the foreseeable future and something that we can fix later on.

Two minute build times with caching are very nice!

PR #20, opened 2023-10-13
Fix C++ Formatting with clang-format

boomanaiden154:
This patch fixes the C++ formatting which was causing the clang-format job to fail. Very minor change, but if we're going to have C++ formatting in CI, we should probably keep it green.


mtrofin (approved):

ondrasej:
LGTM, thanks for fixing it!

PR #23, opened 2023-10-21
Mark the cache/perf counter experiments as x86-64 only.

ondrasej:
The experimental code depends on x86-64 specific intrinsics, and breaks the build on Apple silicon and other non-x86 platforms.

### DIFF
--- a/gematria/experiments/access_pattern_bm/BUILD
+++ b/gematria/experiments/access_pattern_bm/BUILD
@@ -29,6 +29,7 @@ cc_library(
     copts = FEATURE_OPTS,
     target_compatible_with = [
         "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
     ],
 )
 
@@ -54,6 +55,7 @@ cc_library(
     copts = FEATURE_OPTS,
     target_compatible_with = [
         "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
     ],
     deps = ["@com_google_absl//absl/base:core_headers"],
 )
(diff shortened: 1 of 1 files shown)


mtrofin (approved):

boomanaiden154 (approved):
LGTM (not too familiar with the bazel specifics though). There's also the runtime component of these tests (access to performance counters) that cause them to fail at test time in specific configurations, but I'm hoping to get to that at some point in a future patch.

boomanaiden154:
Also looks like this hasn't been rebased on top of a recent commit in `main` which is why there is the double CI jobs/some of the jobs are failing.

ondrasej:
> LGTM (not too familiar with the bazel specifics though). There's also the runtime component of these tests (access to performance counters) that cause them to fail at test time in specific configurations, but I'm hoping to get to that at some point in a future patch.

Good point, perhaps we should limit them to x86-64 & Linux (I don't suppose the perf counter access in googlebenchmark is platform-independent).

> Also looks like this hasn't been rebased on top of a recent commit in main which is why there is the double CI jobs/some of the jobs are failing.

That recent commit is #22? That is what I see as the top of main - maybe the recent commit got lost somewhere?

boomanaiden154:
Nevermind. Wasn't reading things correctly. Half of them are `push` jobs and half of them are `pull_request` jobs because I don't have the triggers set up correctly. I'll get a patch up to fix that.

Given that, the Python formatting errors are probably valid too. Not exactly sure what changed there though to cause them to start failing.

PR #25, opened 2023-10-21
Only run push CI on main branch

boomanaiden154:
Currently the github actions workflow runs whenever something is pushed to a branch or against any pull request. This means that if a branch is pushed to the main google/gematria repository and a PR is opened against google/gematria:main, there will be duplicate jobs, one for the push event and one for the pull request event. This patch fixes this behavior by restricting the push event to the main branch.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,7 +14,12 @@
 
 name: Gematria CI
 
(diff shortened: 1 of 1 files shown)


ondrasej (approved):

ondrasej:
Thanks for cleaning it up!

PR #27, opened 2023-10-30
Add RuntimeAnnotations to supply Gematria models with cache miss information.

virajbshah:
Adds a general framework to add instruction level annotations to the basic block representations, intended to be used to store cache miss frequencies (and later other values, such as branch misprediction related measures).


ondrasej (commented):
Overal, the PR looks good. I'd just go through the naming generailization (what is in the comments in the proto file, and corresponding renames elsewhere).

ondrasej (changes requested):

gematria/basic_block/basic_block.cc
```diff
@@ -229,27 +229,63 @@ std::ostream& operator<<(std::ostream& os, const InstructionOperand& operand) {
   return os;
 }
 
+Annotation::Annotation(const std::string& name, double value)
```
ondrasej, line 232: If we `std::move()`, this should be `std::string name` (no const, no reference). Otherwise, the `std::move()` can't do anything and it must copy the string.

gematria/basic_block/basic_block.h
```diff
@@ -218,6 +218,32 @@ class InstructionOperand {
 
 std::ostream& operator<<(std::ostream& os, const InstructionOperand& operand);
 
+// Represents an annotation holding a value such as some measure/statistic
+// paired with the instruction.
+struct Annotation {
+  // Values default to -1 to have name only annotations.
+  Annotation() : value(-1){};
```
ondrasej, line 225: It's a bit more complex, but maybe we could use a [NaN](https://en.cppreference.com/w/cpp/numeric/math/NAN) for name-only annotations? That way, we could have any valid numerical value as the annotation value.

gematria/basic_block/basic_block_protos.h
```diff
@@ -41,6 +41,12 @@ InstructionOperand InstructionOperandFromProto(
 CanonicalizedOperandProto ProtoFromInstructionOperand(
     const InstructionOperand& operand);
 
+// Creates a annotation data structure from a proto.
```
ondrasej, line 44: Grammar nit: "an annotation"

gematria/basic_block/basic_block_protos.cc
```diff
@@ -112,8 +113,38 @@ void ToRepeatedPtrField(
...
+}
+
+void ToRepeatedPtrField(
+    const std::vector<Annotation>& annotations,
+    google::protobuf::RepeatedPtrField<AnnotationProto>* repeated_field) {
+  repeated_field->Reserve(annotations.size());
+  std::transform(annotations.begin(), annotations.end(),
+                 google::protobuf::RepeatedFieldBackInserter(repeated_field),
+                 ProtoFromAnnotation);
+}
```
ondrasej, line 131: Can we just templatize and reuse the two functions above?

gematria/proto/annotation.proto
```diff
@@ -0,0 +1,27 @@
...
+
+// Contains annotations to supply additional information to the model, such as
+// cache-miss frequencies, or branching related statistics.
+message AnnotationProto {
+  // A name or label for the annotation.
+  string name = 1;
+  // The annotation value, holding information such as measurements or
+  // statistics like event frequency or rate.
+  double value = 2;
+}
```
ondrasej, line 27: Nit: Please add an empty line at the end (otherwise, our internal checker will complain).

gematria/proto/annotation.proto
```diff
@@ -0,0 +1,27 @@
...
+
+package gematria;
+
+// Contains annotations to supply additional information to the model, such as
+// cache-miss frequencies, or branching related statistics.
+message AnnotationProto {
+  // A name or label for the annotation.
+  string name = 1;
+  // The annotation value, holding information such as measurements or
+  // statistics like event frequency or rate.
```
ondrasej, line 25: Please add a comment about name-only annotations.

ondrasej (changes requested):
I went through `basic_block/**` and `proto/**`. I still need to read through `granite/**`.

gematria/basic_block/basic_block.h
virajbshah: I was wondering whether name only annotations should be a thing - I wrote them in for the `proto/**` and `basic_block/**` sides of things, but didn't add support for them from the model's side of things, since that would fit in with the current tensor-overwriting implementation.

gematria/basic_block/basic_block_protos.h
virajbshah: Ah, should've been more careful with the find-and-replacing of the old `RuntimeAnnotation` code :)

virajbshah:
I have made most of the requested changes. I have some questions about whether I should make the changes related to name-only annotations or get rid of them entirely, which I've mentioned in an earlier comment replying to one of the requested changes.

Thanks for your PR review and comments.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -50,6 +52,14 @@ class TokenGraphBuilderModel(graph_builder_model_base.GraphBuilderModelBase):
   READOUT_VARIABLES = 'TokenGraphBuilderModel.readout'
   TASK_READOUT_VARIABLES = 'TokenGraphBuilderModel.task_readout'
 
+  INSTRUCTION_ANNOTATIONS_TENSOR_NAME = (
+      'TokenGraphBuilderModel.instruction_annotations'
+  )
+  NODE_FEATURES_TENSOR_NAME = 'TokenGraphBuilderModel.node_features'
```
ondrasej, line 58: Why do we need this? It doesn't seem to be used anywhere.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -340,3 +358,76 @@ def _create_graph_network_modules(
             residual_connection=options.EnableFeature.BY_FLAG,
         ),
     )
+
+  def _make_batch_feed_dict(self) -> model_base.FeedDict:
+    feed_dict = super()._make_batch_feed_dict()
+    to_feed_instruction_annotations = None
+    if len(self._batch_graph_builder.instruction_annotations) == 0:
```
ondrasej, line 365: Two comments:
1. I suspect that each access to `graph_builder.instruction_annotations` creates a new copy of the map/vectors. That can be very expensive, and it's best avoided.
2. The "canonical" way to check whether a container is empty in Python is to use the implicit boolean conversion.

```
instruction_annotations = self._batch_graph_builder.instruction_annotations
if not instruction_annotations:
  ...
```

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -340,3 +358,76 @@ def _create_graph_network_modules(
...
+      to_feed_instruction_annotations = np.zeros((0, 0), dtype=self.dtype)
+    else:
+      to_feed_instruction_annotations = np.array(
+          list(self._batch_graph_builder.instruction_annotations.values()),
```
ondrasej, line 369: It would be nice if we could just `reshape()` the data provided from the graph builder, to avoid too much Python processing.

ondrasej, line 369: This may be dangerous: Python dicts do have a stable iteration order (iteration is in the insertion order), but this doesn't hold for C++ unordered containers, and we might be getting the values in a random order each time we run the code.

I'd prefer manually sorting columns of the tensor by a provided list of annotations (perhaps even in the C++ code, for more efficiency).

ondrasej (changes requested):

gematria/granite/python/graph_builder.cc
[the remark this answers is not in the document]
virajbshah: I have been putting that off since doing that in a way that is consistent with the existing tests in `graph_builder_test.py` would require adding annotation data to all of the blocks in `basic_blocks_with_throughput.pbtxt` (or otherwise guaranteeing at least some of the sampled blocks used within the tests have annotations, maybe by having an `annotated_basic_blocks_with_throughput.pbtxt` or similar).

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -335,6 +338,37 @@ llvm::Expected<std::string> GetNodeTokenAtIndex(
   return node_token_list[token_index];
 }
 
+// Extracts the set of annotation names from the model. This should be a Const
+// tensor, and as such, it should be readable without providing any inputs.
+// Returns an error when the annotation names tensor is not found or when it is
+// not readable.
```
ondrasej, line 345: Nit: Please add a TODO (for me or for you) to extract most of the logic from here and the GetNodeTokenList() to a single function.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -373,6 +407,12 @@ GraphBuilderModelInference::FromTfLiteModel(
         llvm::errc::invalid_argument,
         "The special token index tensor could not be read");
   }
+
+  // Get the set of annotation names used by the model.
+  llvm::Expected<std::set<std::string>> annotation_names =
+      GetAnnotationNames(**interpreter);
```
ondrasej, line 414: I'm a bit worried here - requiring that the list of annotations is always present will invalidate all the existing tflite files (not that there are too many). Maybe we should have a flag allowing the load the older models that do not have annotations.

ondrasej (commented):
I've added a few more comments, but otherwise it looks good.

virajbshah:
This PR is being split into 4 separate PRs:
 * #92 
 * #93 
 * #94 
 * And an upcoming PR including the C++ inference API changes from this PR with some improvements.

virajbshah:
The 4th PR has now been opened as well:
 * #98

PR #28, opened 2023-11-13
Add script to generate input code snippets for llvm-exegesis

9Tempest:
Added a script `dataset/convert_bhive_to_llvm_exegesis_input.cc`  that takes a bhive csv file containing a sort of x86 basic blocks in hex format and a `llvm-exegesis` template that defines initial register values (`dataset/llvm-exegesis_wrapper.S` is used by default) as input. It will output code snippets that can be executed by `llvm-exegesis` in target output_dir. Each code snippet(output file) will contain one x86 basic block.

### How to run it?
```bash
# build the latest repo
bazel build ...

# create an output directory
mkdir output

# run script
./bazel-bin/gematria/datasets/convert_bhive_to_llvm_exegesis_input --bhive_csv=gematria/testing/testdata/test.csv  --output_dir=output
```


gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+      std::cerr << "could not parse: " << hex << "\n";
+      return 3;
+    }
+
+    // for each line, find the accessed addresses & disassemble instructions
+    const auto& bytes = bytes_or.value();
+    auto addrs_or = gematria::FindAccessedAddrs(bytes);
+    auto proto = bhive_importer.BasicBlockProtoFromMachineCode(bytes);
+    if (addrs_or.ok() && proto.ok()) {
+      // TODO: attach those addresses to input that can be executed by
```
boomanaiden154, line 100: It seems like this TODO is done as you're adding the memory annotations shortly afterwards (line 117)?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+      output_file << template_content;
+
+      // Append memory annotations
+      auto addrs = addrs_or.value();
+      for (const auto& addr : addrs.accessed_blocks) {
+        std::string mem_name =
+            LLVM_EXEGESIS_MEM_NAME_PREFIX + std::to_string(mem_counter++);
+        output_file << LLVM_EXEGESIS_MEM_DEF_PREFIX << mem_name << " "
+                    << addrs.block_size << " " << INITIAL_MEM_VALUE << "\n";
+        output_file << LLVM_EXEGESIS_MEM_MAP_PREFIX << mem_name << " "
```
boomanaiden154, line 122: The memory name here should probably be fixed and all the mappings should point to a single definition. Using multiple definitions means that the BB will access multiple pages, which means that there might be a significant number of cache misses due to everything not fitting in the L1 dcache.

gematria/testing/testdata/test.csv
```diff
@@ -0,0 +1,154 @@
+4883c2014883fa40,0.00001339
```
boomanaiden154, line 1: It doesn't seem like this test data is wired up into a test anywhere? Not sure what the best way to wire it up would be, but it would be good to have some test coverage for this.

gematria/datasets/llvm-exegesis_warpper.S
```diff
@@ -0,0 +1,32 @@
+# LLVM-EXEGESIS-DEFREG RAX 10000
```
boomanaiden154, line 1: Having this produced/embedded in the executable seems like it would be cleaner than loading it in at runtime? Not sure what @ondrasej thinks about that though.

gematria/datasets/llvm-exegesis_warpper.S
```diff
@@ -0,0 +1,32 @@
...
+# LLVM-EXEGESIS-DEFREG RBX 10000
+# LLVM-EXEGESIS-DEFREG RCX 10000
+# LLVM-EXEGESIS-DEFREG RDX 10000
+# LLVM-EXEGESIS-DEFREG RSI 10000
```
boomanaiden154, line 5: It looks like you're omitting `RSP` here due to the issues that we discussed over slack? It should be preserved and set properly by `llvm-exegesis`, so it's currently a bug (thinking about it more now, some of the register values are loaded in from the stack and there's no register saving there, so it's most likely getting clobbered). I'll try and put up a patch soonish. Can you add a note/comment somewhere that it's not being set currently due to that reason?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+    std::string_view hex = std::string_view(line).substr(0, comma_index);
+    auto bytes_or = gematria::ParseHexString(hex);
+    if (!bytes_or.has_value()) {
+      std::cerr << "could not parse: " << hex << "\n";
+      return 3;
+    }
+
+    // for each line, find the accessed addresses & disassemble instructions
+    const auto& bytes = bytes_or.value();
+    auto addrs_or = gematria::FindAccessedAddrs(bytes);
```
boomanaiden154, line 97: Might be good to add a comment here that this will only get the first segfault address. Looking at the TODO near https://github.com/google/gematria/blob/a5bac1d2237d95f901264e3944ba8879244ac489/gematria/datasets/find_accessed_addrs.cc#L327 It doesn't seem like it should be too difficult to get that all wired up, but it does need to be done (ideally with fairly extensive testing).

boomanaiden154 (commented):
Some initial comments on what I've observed. One more question:
* Is `llvm-exegesis` successfully able to execute all the blocks in the test data after being run through this script?

And one more comment:
* Doing this by generating files for `llvm-exegesis` looks good to me for now, but we can also call into the `llvm-exegesis` API directly by linking in LLVM and that might offer some more flexibility in the future as we scale to a lot more data.

gematria/testing/testdata/test.csv
9Tempest: Yeah... I randomly copied pasted from [one bhive benchmark](https://github.com/GranLte/bhive/blob/master/benchmark/sources/sqlite.csv) for testing...

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
9Tempest: Oh, forgot to remove... Will remove shortly!

9Tempest: Done

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
9Tempest: Done

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,147 @@
...
+
+      // Multiple mappings can point to the same definition
+      if (addrs.accessed_blocks.size() > 0) {
+        output_file << LLVM_EXEGESIS_MEM_DEF_PREFIX
+                    << LLVM_EXEGESIS_MEM_NAME_PREFIX << " " << addrs.block_size
+                    << " " << INITIAL_MEM_VALUE << "\n";
+      }
+      for (const auto& addr : addrs.accessed_blocks) {
+        output_file << LLVM_EXEGESIS_MEM_MAP_PREFIX
+                    << LLVM_EXEGESIS_MEM_NAME_PREFIX << " " << std::dec << addr
```
boomanaiden154, line 126: This will need to be adjusted after https://github.com/llvm/llvm-project/pull/72620 lands.

boomanaiden154 (commented):
One additional comment as I'm making more changes in upstream `llvm-exegesis`.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+  if (bhive_filename.empty()) {
+    std::cerr << "Error: --bhive_csv is required\n";
+    return 1;
+  }
+
+  std::string output_dir = absl::GetFlag(FLAGS_output_dir);
+  if (output_dir.empty()) {
+    std::cerr << "Error: --output_dir is required\n";
+    return 1;
+  }
```
ondrasej, line 60: Same here.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+    std::string, output_dir, "",
+    "Directory containing output files that can be executed by llvm-exegesis");
+ABSL_FLAG(std::string, exegesis_template,
+          "gematria/datasets/llvm-exegesis_warpper.S",
+          "Template file for llvm-exegesis");
+
+int main(int argc, char* argv[]) {
+  absl::ParseCommandLine(argc, argv);
+
+  std::string bhive_filename = absl::GetFlag(FLAGS_bhive_csv);
```
ondrasej, line 50: This is a bit nitpicky, but
```
const std::string bhive_filename = absl::GetFlag(FLAGS_bhive_csv);
```

I'd generally prefer adding `const` wherever possible. It helps avoiding unintended modifications of data, and also helps with reasoning about what can change when and where.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+  std::ifstream bhive_csv_file(bhive_filename);
+  for (std::string line; std::getline(bhive_csv_file, line);) {
+    auto comma_index = line.find(',');
+    if (comma_index == std::string::npos) {
+      std::cerr << "Invalid CSV file: no comma in line '" << line << "'\n";
+      return 2;
+    }
+
+    std::string_view hex = std::string_view(line).substr(0, comma_index);
+    auto bytes_or = gematria::ParseHexString(hex);
```
ondrasej, line 88: The general recommendation is to not use an "_or" suffix, and just treat it similarly to smart pointers (use `->` for member access or `*` for dereferencing).

See https://abseil.io/tips/181 for more details. The tip is about `absl::StatusOr`, but a lot of it applies also to `std::optional`.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+    if (!bytes_or.has_value()) {
+      std::cerr << "could not parse: " << hex << "\n";
+      return 3;
+    }
+
+    // for each line, find the accessed addresses & disassemble instructions
```
ondrasej, line 94: Very nit: Comments should be formatted like sentences (start with a capital letter, end with a period).

(see https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar for more thoughts on that)

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
ondrasej: Please change the header of the TODO to `TODO(9Tempest):`. In general, TODOs should be tagged with the person responsible, or with a link to an issue. Alternatively, we can create an issue here in the repo to track this.

See https://google.github.io/styleguide/cppguide.html#TODO_Comments.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+    const auto& bytes = bytes_or.value();
+    auto addrs_or = gematria::FindAccessedAddrs(bytes);
+    auto proto = bhive_importer.BasicBlockProtoFromMachineCode(bytes);
+    if (addrs_or.ok() && proto.ok()) {
+      // TODO: attach those addresses to input that can be executed by
+      // llvm-exegesis Create output file path
```
ondrasej: This looks like two comments accidentally merged into one.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,143 @@
...
+    } else if (proto.ok()) {
+      std::cerr << "Failed to find addresses for block '" << hex
+                << "': " << addrs_or.status() << "\n";
+      std::cerr << "Block disassembly:\n";
+      for (const auto& instr : proto->machine_instructions()) {
+        std::cerr << "\t" << instr.assembly() << "\n";
+      }
+    } else {
+      std::cerr << "Failed to disassemble block '" << hex << "\n";
+    }
```
ondrasej, line 145: I'd prefer to put this to the beginning of the function, with a `continue` if there is a non-critical failure. Also, we should print the status to make debugging easier.

```
if (!proto.ok()) {
  std::cerr << "Failed to disassemble block '" << hex << "': " << proto.status() << "\n";
  continue;
}
// ...
if (!addrs.ok()) { 
  std::cerr << "Failed to determine used addresses for block '" << hex << "': "
<< addrs.status() << " \n";
  // ...print disassembly...
  continue;
}
```

ondrasej (changes requested):
First of all, thanks for working on this!

Second, I'm sorry for taking so long on the review. Please ping me if I'm blocking you next time.

Overall, this looks good, I've added comments mainly on style (and compatibility with the Google Style guide). Please let me know if you have any questions.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Patch got closed, so this should be fine for now.

9Tempest:
> First of all, thanks for working on this!
> 
> Second, I'm sorry for taking so long on the review. Please ping me if I'm blocking you next time.
> 
> Overall, this looks good, I've added comments mainly on style (and compatibility with the Google Style guide). Please let me know if you have any questions.

Thanks for the review! I agree that we should get the registers from llvm register info classes such that it can be easily extended to other ISAs. Will work on revising the code style and using register info classes these days

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,153 @@
...
+    // Filter out subregisters. Only include the super register.
+    bool is_sub_reg = false;
+    for (auto SuperReg : MRI.superregs(I)) {
+      if (MRI.isSubRegister(SuperReg, I)) {
+        is_sub_reg = true;
+        break;
+      }
+    }
+    if (is_sub_reg) {
+      continue;
```
boomanaiden154, line 75: This needs some adjustment. It returns all possible registers for X86 as it doesn't take into account the current CPU features. So you'll end up with all the vector registers and even GPRs that aren't on currently released processors (as APX is supported in upstream LLVM now).

It'll probably be a lot easier to just look through specific register classes.  `X86::GR64_NOREX2` and `X86::VR128` should get what you need. The following two loops shown an example of how to iterate over the classes to get the names.

```c++
for(unsigned i = 0; i < MRI.getRegClass(X86::GR64_NOREX2RegClassID).getNumRegs(); ++i) {
  std::cout << MRI.getName(MRI.getRegClass(X86::GR64_NOREX2RegClassID).getRegister(i)) << "\n";
}

for(unsigned i = 0; i < MRI.getRegClass(X86::VR128RegClassID).getNumRegs(); ++i) {
  std::cout << MRI.getName(MRI.getRegClass(X86::VR128RegClassID).getRegister(i)) << "\n";
}
```

At some point we'll probably want to wire this into a `TargetRegisterInfo` so we can actually query information on what vectors are available for the hardware we're running on, but for now just iterating through the classes should be fine.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,155 @@
...
+
+  std::string register_defs_lines;
+  const std::unique_ptr<gematria::LlvmArchitectureSupport> llvm_support =
+      gematria::LlvmArchitectureSupport::X86_64();
+  const llvm::MCRegisterInfo& MRI = llvm_support->mc_register_info();
+
+  // Iterate through all general purpose registers and vector registers
+  // and add them to the register definitions.
+  for (unsigned i = 0;
+       i < MRI.getRegClass(llvm::X86::GR64_NOREXRegClassID).getNumRegs(); ++i) {
```
boomanaiden154, line 71: The correct register class here for x86-64 should be `GR64_NOREX2RegClassID`. Just using `GR64_NOREXRegClassID` only pulls in the first eight x86-64 registers and omits r9-r15.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,155 @@
...
+    llvm::StringRef reg_name =
+        MRI.getName(MRI.getRegClass(llvm::X86::GR64_NOREXRegClassID).getRegister(i));
+    register_defs_lines += std::string(kRegDefPrefix) + std::string(reg_name) +
+                           " " + std::to_string(kInitialRegVal) + "\n";
```
boomanaiden154, line 75: The register value here (and below when iterating through the vector registers) is interpreted as a hex value. So we should probably change the constant above to `0x10000` and then move this all over to a `stringstream` and do `std::hex << kInitialRegValue << std::dec` (or whatever else to get the hex string from the integer).

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,155 @@
...
+      return 4;
+    }
+
+    // Write the register definition lines into the output file.
+    output_file << register_defs_lines;
+
+    // Multiple mappings can point to the same definition.
+    if (addrs->accessed_blocks.size() > 0) {
+      output_file << kMemDefPrefix << kMemNamePrefix << " " << addrs->block_size
+                  << " " << kInitialMemVal << "\n";
```
boomanaiden154, line 141: `kInitialMemVal` is a hex value, so we should convert the integer to a hex value here rather than just printing the decimal number.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -60,26 +60,35 @@ int main(int argc, char* argv[]) {
...
+  }
   std::string register_defs_lines;
   const std::unique_ptr<gematria::LlvmArchitectureSupport> llvm_support =
       gematria::LlvmArchitectureSupport::X86_64();
   const llvm::MCRegisterInfo& MRI = llvm_support->mc_register_info();
 
   // Iterate through all general purpose registers and vector registers
   // and add them to the register definitions.
+  // TODO(9Tempest): Change GR64_NOREXRegClassID to GR64_NOREX2RegClassID to
+  // include r9-r15
```
boomanaiden154, line 78: I don't think it's really correct to say that we're including r9-r15 by switching the register class. I'd instead writing something like:
```c++
// TODO(9Temptest): Change GR64_NOREXRegClassID to GR64_NOREX2RegClassID when the LLVM version
// is bumped to avoid including the new APX GPRs (r16-r31) that have recently been added to LLVM.
```

gematria/testing/testdata/test.csv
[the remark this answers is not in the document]
boomanaiden154: @9Tempest I'm thinking we can probably just delete this file for the initial patch to avoid the licensing issues? We won't have any automated test coverage, but we wouldn't even if the file was checked in.

boomanaiden154 (approved):
Two minor comments still outstanding that should be easy to address, otherwise this LGTM. There are definitely still some improvements that we'll want to make in the future, but this is fine in it's initial state and the improvements can be made in tree.

ondrasej (changes requested):
I've added a few more comments. The variable naming style is important. The string conversions would be nice to have, but we can clean them up later if needed.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -60,43 +60,50 @@ int main(int argc, char* argv[]) {
...
-  }
-  std::string initialMemValStr;
-  {
-    // initialize initialMemValStr with kInitialMemValue in hex
+    // initialize initial_mem_val_str with kInitialMemValue in hex
     std::stringstream ss;
     ss << std::hex << kInitialMemVal << std::dec;
-    ss >> initialMemValStr;
+    ss >> initial_mem_val_str;
   }
```
ondrasej, line 72: Is this part still needed? This should be handled by the `ConvertHexTostring()` on the line above, no?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -60,43 +60,50 @@ int main(int argc, char* argv[]) {
...
-       i < MRI.getRegClass(llvm::X86::GR64_NOREXRegClassID).getNumRegs(); ++i) {
-    llvm::StringRef reg_name = MRI.getName(
-        MRI.getRegClass(llvm::X86::GR64_NOREXRegClassID).getRegister(i));
-    register_defs_lines += std::string(kRegDefPrefix) + std::string(reg_name) +
-                           " " + initialRegValStr + "\n";
+       i < reg_info.getRegClass(llvm::X86::GR64_NOREXRegClassID).getNumRegs();
+       ++i) {
+    llvm::StringRef reg_name = reg_info.getName(
+        reg_info.getRegClass(llvm::X86::GR64_NOREXRegClassID).getRegister(i));
+    register_defs_lines = llvm::Twine(register_defs_lines)
```
ondrasej, line 88: This will still make mroe copying, I'd keep the `+=` here, but use a twine for the concats, e.g.
```
register_defs_lines += llvm::Twine(kRegDefPrefix).concat(reg_name).concat(" ") ...;
```

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -60,43 +60,50 @@ int main(int argc, char* argv[]) {
...
   for (unsigned i = 0;
-       i < MRI.getRegClass(llvm::X86::VR128RegClassID).getNumRegs(); ++i) {
-    llvm::StringRef reg_name =
-        MRI.getName(MRI.getRegClass(llvm::X86::VR128RegClassID).getRegister(i));
-    register_defs_lines += std::string(kRegDefPrefix) + std::string(reg_name) +
-                           " " + initialRegValStr + "\n";
+       i < reg_info.getRegClass(llvm::X86::VR128RegClassID).getNumRegs(); ++i) {
+    llvm::StringRef reg_name = reg_info.getName(
+        reg_info.getRegClass(llvm::X86::VR128RegClassID).getRegister(i));
+    register_defs_lines = llvm::Twine(register_defs_lines)
```
ondrasej, line 100: Same here.

ondrasej (changes requested):
I've added a few more comments, but otherwise it looks good to go.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -0,0 +1,176 @@
...
+constexpr uint64_t kInitialRegVal = 0x10000;
+constexpr uint64_t kInitialMemVal = 0x7FFFFFFF;
+constexpr std::string_view kRegDefPrefix = "# LLVM-EXEGESIS-DEFREG ";
+constexpr std::string_view kMemDefPrefix = "# LLVM-EXEGESIS-MEM-DEF ";
+constexpr std::string_view kMemMapPrefix = "# LLVM-EXEGESIS-MEM-MAP ";
+constexpr std::string_view kMemNamePrefix = "MEM";
+
+namespace {
+unsigned int file_counter = 0;
+}
```
ondrasej, line 41: I'd keep this as a local variable in `main()`. It's only really used from there, so there's no need to make it global.

ondrasej (changes requested):
Last comment (I hope :))

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
ondrasej: I've merged the PR as is to expedite it, and I'll fix this myself afterwards.

ondrasej:
Thanks for the contribution! I'm sorry the reviews took so long.

PR #30, opened 2023-12-02
Add exegesis BB annotator

boomanaiden154:
This patch adds in an alternative to the existing find_accessed_addrs infrastructure that uses llvm-exegesis as a backend. We believe this will more closely match the execution environment of the benchmarking environment and should avoid some of the pitfalls of the existing memory annotation infrastructure (at the current expense of speed).

### DIFF
--- a/gematria/datasets/find_accessed_addrs_exegesis.cc
+++ b/gematria/datasets/find_accessed_addrs_exegesis.cc
@@ -14,14 +14,18 @@
 
 #include "gematria/datasets/find_accessed_addrs_exegesis.h"
 
-#include "BenchmarkRunner.h"
-#include "LlvmState.h"
-#include "Target.h"
-#include "TargetSelect.h"
+// Use the absolute path for headers from llvm-exegesis as there is no
+// canonical include path within LLVM as they are not properly exposed through
+// a library and could potentially be confused with other LLVM includes.
+
 #include "X86.h"
 #include "X86InstrInfo.h"
 #include "X86RegisterInfo.h"
 #include "gematria/llvm/disassembler.h"
+#include "llvm/tools/llvm-exegesis/lib/BenchmarkRunner.h"
+#include "llvm/tools/llvm-exegesis/lib/LlvmState.h"
+#include "llvm/tools/llvm-exegesis/lib/Target.h"
+#include "llvm/tools/llvm-exegesis/lib/TargetSelect.h"
 
 using namespace llvm;
 using namespace llvm::exegesis;
--- a/gematria/datasets/find_accessed_addrs_exegesis.cc
+++ b/gematria/datasets/find_accessed_addrs_exegesis.cc
@@ -33,17 +33,26 @@ using namespace llvm::exegesis;
 namespace gematria {
 
 ExegesisAnnotator::ExegesisAnnotator(
-    LlvmArchitectureSupport &ArchSup, LLVMState &ExegesisState,
-    std::unique_ptr<BenchmarkRunner> BenchRunner,
+    LLVMState &ExegesisState, std::unique_ptr<BenchmarkRunner> BenchRunner,
     std::unique_ptr<const SnippetRepetitor> SnipRepetitor)
-    : ArchSupport(ArchSup),
-      MCPrinter(ArchSupport.CreateMCInstPrinter(0)),
-      State(ExegesisState),
+    : State(ExegesisState),
       Runner(std::move(BenchRunner)),
-      Repetitor(std::move(SnipRepetitor)) {}
+      Repetitor(std::move(SnipRepetitor)) {
+  MachineContext = std::make_unique<MCContext>(
+      State.getTargetMachine().getTargetTriple(),
+      State.getTargetMachine().getMCAsmInfo(), &State.getRegInfo(),
+      &State.getSubtargetInfo());
+  MachineDisassembler.reset(
+      State.getTargetMachine().getTarget().createMCDisassembler(
+          State.getSubtargetInfo(), *MachineContext));
+  MachinePrinter.reset(State.getTargetMachine().getTarget().createMCInstPrinter(
+      State.getTargetMachine().getTargetTriple(), 0,
+      *State.getTargetMachine().getMCAsmInfo(), State.getInstrInfo(),
+      State.getRegInfo()));
+}
 
 Expected<std::unique_ptr<ExegesisAnnotator>> ExegesisAnnotator::create(
-    LlvmArchitectureSupport &ArchSup, LLVMState &ExegesisState) {
+    LLVMState &ExegesisState) {
   // Initialize the supported Exegesis targets. Currently we only support X86.
   InitializeX86ExegesisTarget();
 
@@ -62,18 +71,16 @@ Expected<std::unique_ptr<ExegesisAnnotator>> ExegesisAnnotator::create(
       SnippetRepetitor::Create(Benchmark::RepetitionModeE::Duplicate,
                                ExegesisState);
 
-  return std::unique_ptr<ExegesisAnnotator>(
-      new ExegesisAnnotator(ArchSup, ExegesisState, std::move(*RunnerOrErr),
-                            std::move(SnipRepetitor)));
+  return std::unique_ptr<ExegesisAnnotator>(new ExegesisAnnotator(
+      ExegesisState, std::move(*RunnerOrErr), std::move(SnipRepetitor)));
 }
 
 Expected<AccessedAddrs> ExegesisAnnotator::findAccessedAddrs(
     ArrayRef<uint8_t> BasicBlock) {
   Expected<std::vector<DisassembledInstruction>> DisInstructions =
-      DisassembleAllInstructions(
-          ArchSupport.mc_disassembler(), ArchSupport.mc_instr_info(),
-          ArchSupport.mc_register_info(), ArchSupport.mc_subtarget_info(),
-          *MCPrinter, 0, BasicBlock);
+      DisassembleAllInstructions(*MachineDisassembler, State.getInstrInfo(),
+                                 State.getRegInfo(), State.getSubtargetInfo(),
+                                 *MachinePrinter, 0, BasicBlock);
 
   if (!DisInstructions) return DisInstructions.takeError();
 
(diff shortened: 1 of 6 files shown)


gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -0,0 +1,152 @@
...
+    // If we don't have any errors executing the snippet, we executed the
+    // snippet successfully and thus have all the needed memory annotations.
+    if (BenchmarkResults) break;
+
+    auto ResultError = BenchmarkResults.takeError();
+
+    if (!ResultError.isA<SnippetCrash>()) return std::move(ResultError);
+
+    handleAllErrors(std::move(ResultError), [&](SnippetCrash &CrashInfo) {
+      if (CrashInfo.GetCrashAddress() == 0) return;
```
boomanaiden154, line 134: This probably needs to turn into an actual crash. We can map pages under the 64kb limit in some contexts, but it's generally not an encouraged practice. Will have to see the practical implications though.

This case currently happens transiently due to the restartable sequence not getting disabled in glibc due to some weird preprocessor stuff happening in Bazel. Will investigate it and I think it will probably end up needing a patch on the LLVM side. Hacky fix that I've been experimenting with is to adjust some of the preprocessor logic in the head of `llvm/tools/llvm-exegesis/lib/BenchmarkRunner.cpp`.

boomanaiden154: Ah. It's because `HAVE_BUILTIN_THREAD_POINTER` is necessary for the preprocessor check to pass and is defined by CMake, but isn't defined by the Bazel build. Will post a patch to get that fixed.

boomanaiden154:
This should be pretty much good to go. https://github.com/llvm/llvm-project/pull/74574 still needs to land for the comment mentioned above, but there shouldn't be any other issues.

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Fixed in https://github.com/llvm/llvm-project/pull/74574.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -0,0 +1,151 @@
...
+
+#include "gematria/datasets/find_accessed_addrs_exegesis.h"
+
+#include "BenchmarkRunner.h"
+#include "LlvmState.h"
+#include "Target.h"
+#include "TargetSelect.h"
+#include "X86.h"
+#include "X86InstrInfo.h"
+#include "X86RegisterInfo.h"
```
ondrasej, line 23: These look a bit suspicious - I guess we have some `-I` somewhere that pulls more directories than it should. But even then we should be using llvm-absolute paths or gematria-absolute paths for the includes.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -0,0 +1,151 @@
...
+#include "gematria/llvm/disassembler.h"
+
+using namespace llvm;
+using namespace llvm::exegesis;
```
ondrasej, line 27: We should follow either the Google style guide (which disallows `using namespace` or converge on the LLVM style guide, but that would require different naming for methods in the class (and IIRC, member variables which should not end with an underscore).

I'm fine with both styles (most of the code uses the Google style now, but I think we'll need to move to the LLVM style eventually, so it would be fine to use it, as long as it is used consistently and for larger blocks of code).

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -0,0 +1,151 @@
...
+        "Failed to initialize libpfm",
+        std::make_error_code(std::errc::invalid_argument));
+
+  auto RunnerOrErr = State_.getExegesisTarget().createBenchmarkRunner(
+      Benchmark::Latency, State_, BenchmarkPhaseSelectorE::Measure,
+      BenchmarkRunner::ExecutionModeE::SubProcess, 1, Benchmark::Min);
+
+  if (!RunnerOrErr) return RunnerOrErr.takeError();
+
+  std::unique_ptr<const SnippetRepetitor> Repetitor_ =
```
ondrasej, line 57: No trailing underscore in the variable name.

gematria/datasets/find_accessed_addrs_test.cc
```diff
@@ -16,22 +16,26 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <iostream>
 #include <memory>
 #include <string>
 #include <string_view>
 
+#include "TargetSelect.h"
```
ondrasej, line 24: This looks suspicious, similar to the includes in the .cc file.

gematria/datasets/find_accessed_addrs_test.cc
```diff
@@ -161,5 +179,24 @@ TEST_F(FindAccessedAddrsTest, AccessFromRegister) {
                                  ElementsAre(0x10000, 0x20000))));
 }
 
+TEST_F(FindAccessedAddrsTest, ExegesisNoAccess) {
+  auto addrs_or_err = FindAccessedAddrsExegesis(R"asm(
+    mov r11, r12
+  )asm");
+  ASSERT_TRUE(static_cast<bool>(addrs_or_err));
```
ondrasej, line 186: Does LLVM have a gmock matcher for `llvm::Expected` similar to `IsOkAndHolds`?

I'd expect it to have one, and if not, it might be worth adding one (eventually, not in this PR). This is a very convenient way to write tests.

gematria/datasets/find_accessed_addrs_test.cc
ondrasej, line 1: I'm not sure if we should be keeping the tests in the same file. That said, we should eventually merge both implementations (and perhaps move them to the LLVM repo, next to llvm-exegesis, but that's not something we need to worry about right now).

If we keep the tests in the same file, it would be nice to be able to use tests in both implementations, so that we don't need to write them twice. [Parameterized tests](https://google.github.io/googletest/reference/testing.html#TEST_P) might help with this.

ondrasej (changes requested):
Nice, thanks for working on this!

I have a bunch of changes in the existing tool that I need to push, and Owen keeps working on it. We need to eventually merge the two and keep the best from both versions (and then we should start thinking about moving it to the LLVM repo, because IMO it would make most sense next to llvm-exegesis).

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: The top four includes are from Exegesis. Given that Exegesis doesn't expose any headers in `llvm/include` and they're just inside the `llvm/tools/llvm-exegesis/lib` directory where all the source code lives, they don't really have an absolute path.

The X86 headers are also in a `llvm/*/lib` directory (not entirely sure why) and the canonical way to include them across all of LLVM is the way I've done it here.

gematria/datasets/find_accessed_addrs_test.cc
boomanaiden154: This is also an `llvm-exegesis` include.

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Good point. I prefer the LLVM Style, so that's mostly what I've written this in. I kept the method names with slightly different naming to be consistent with `FindAccessedAddrs`, but staying completely in the LLVM Style makes sense and I've updated the patch.

I don't think there's anything explicit about disallowing underscores within the LLVM Style guide, but I've renamed things to get rid of it as it definitely isn't canonical within LLVM.

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Fixed

gematria/datasets/find_accessed_addrs_test.cc
boomanaiden154: I've split this into a separate test file. Some of the really basic tests might end up being the same (like one access/two access tests), but I suspect a lot of the regression tests will end up being quite a bit different if they end up popping up given how different the underlying implementations are.

The big test is probably the randomized one which is already disabled for being slow with the existing annotator, and the exegesis annotator is currently significantly slower.

gematria/datasets/find_accessed_addrs_test.cc
boomanaiden154: It does, specifically `ASSERT_THAT_EXPECTED`, but linking against the library that provides it, `@llvm-project//llvm:TestingSupport` currently is causing test failures during cleanup due to a pthread error. I've left a TODO for the time being.

ondrasej:
> > I have a bunch of changes in the existing tool that I need to push, and Owen keeps working on it. We need to eventually merge the two and keep the best from both versions (and then we should start thinking about moving it to the LLVM repo, because IMO it would make most sense next to llvm-exegesis).
> 
> Yep. Consolidating on something eventually would be good. It might not be this. This was mostly just something I threw together as when I was experimenting with the current `FindAccessedAddrs` implementation, it was finding addresses that would then segfault in Exegesis, and I figured directly using Exegesis would fix any discrepancies there during experimentation.
> 
> Owen's annotator is significantly faster. The Exegesis annotator is pretty slow. I haven't profiled it yet, but I suspect at least some of the slowness is inherent to the design. Everything has to get emitted to textual assembly, reparsed, reassembled, and then executed. Something to fix this would be nice, but I don't think it would be trivial.
> 
> Upstreaming definitely is something to look into at some point. We'll have to see where this ends up going and what ends up scaling well.

SGTM. Let me upstream that first.

boomanaiden154:
I've updated the patch to fix the includes in the various places where `llvm-exegesis` includes are used and added a comment where they are noting why we're only using the absolute paths there.

ondrasej (approved):

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: For the Exegesis includes I did them relative to the root of the LLVM monorepo. I've left the X86 includes as-is (for now) given that's how they're included in LLVM.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "X86RegisterInfo.h"
+#include "gematria/llvm/disassembler.h"
+
+using namespace llvm;
+using namespace llvm::exegesis;
+
+namespace gematria {
+
+ExegesisAnnotator::ExegesisAnnotator(
+    LlvmArchitectureSupport &ArchSup, LLVMState &ExegesisState,
```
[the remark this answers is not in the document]
boomanaiden154: Good point. I've switched to only using `LLVMState` and storing a couple additional necessary components that would normally be held by `LlvmArchitectureSupport` so that we can keep everything consolidated to one (at least within the class).

PR #32, opened 2024-01-12
Set Bazel version in CI

boomanaiden154:
Currently the Bazel build is failing due to the CI trying to pull in bazel v7.0.0. This patch pins the bazel version to v6.4.0 which gets the CI green again, allowing further time to work on performing the necessary migration steps to get everything working properly with v7.0.0.


boomanaiden154:
We need to convert to Bzlmod at some point, but this keeps things working for now.

ondrasej:
Thanks for looking into this!

PR #33, opened 2024-01-27
Make exegesis annotator work with mappings that aren't page aligned

boomanaiden154:
This patch makes the exegesis annotator work with mappings that are not page aligned. Currently if a segfault occurs at an address that isn't page aligned, Exegesis will try and map an address at that address and then the mmap call will silently fail as there is no error handling wired up for that and it will segfault again, thus creating a loop within the annotator.

### DIFF
--- a/gematria/datasets/find_accessed_addrs_exegesis.cc
+++ b/gematria/datasets/find_accessed_addrs_exegesis.cc
@@ -146,8 +146,9 @@ Expected<AccessedAddrs> ExegesisAnnotator::findAccessedAddrs(
     handleAllErrors(std::move(std::get<0>(BenchmarkResultOrErr)),
                     [&](SnippetSegmentationFault &CrashInfo) {
                       MemoryMapping MemMap;
-                      uintptr_t MapAddress =
-                          (CrashInfo.getAddress() / 4096) * 4096;
+                      // Zero out the last twelve bits of the address to align
+                      // the address to a page boundary.
+                      uintptr_t MapAddress = (CrashInfo.getAddress() & ~0xfff);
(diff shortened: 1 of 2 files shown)


mtrofin:
> here is no error handling wired up for that

Should there be one wired up then?

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -146,7 +146,9 @@ Expected<AccessedAddrs> ExegesisAnnotator::findAccessedAddrs(
     handleAllErrors(std::move(std::get<0>(BenchmarkResultOrErr)),
                     [&](SnippetSegmentationFault &CrashInfo) {
                       MemoryMapping MemMap;
-                      MemMap.Address = CrashInfo.getAddress();
+                      uintptr_t MapAddress =
```
mtrofin, line 149: You could just clear the last 12 bits of `CrashInfo.getAddress()`, that would be (to me) more clear I think.

mtrofin (approved):
lgtm, with a suggestion

boomanaiden154:
> Should there be one wired up then?

Yes. It's been on my todo list for a while. It's not trivial to add though as the mmap calls that potentially fail are in the assembled benchmark snippet which is manually created assembly using the LLVM APIs. So it's possible to do a conditional jump if `%rax` after `syscall` isn't the expected value and then exit from there, but I haven't looked at implementing it yet.

This is also less of an issue if assertions are enabled as https://github.com/llvm/llvm-project/pull/75554 adds an assertion that triggers when this case occurs (which I did hit after bumping the LLVM version).

PR #35, opened 2024-01-27
Add max_bb_count to exegesis converter

boomanaiden154:
This patch adds a max_bb_count flag to the exegesis converter which makes it easier to generate only a subset of basic blocks from a large CSV for testing purposes. This reduces the need for manually splitting CSVs at the cost of little additional complexity.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -14,6 +14,7 @@
 
 #include <fstream>
 #include <iostream>
+#include <limits>
 #include <memory>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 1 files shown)


gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -102,7 +105,10 @@ int main(int argc, char* argv[]) {
   gematria::BHiveImporter bhive_importer(&canonicalizer);
 
   std::ifstream bhive_csv_file(bhive_filename);
+  const unsigned max_bb_count = absl::GetFlag(FLAGS_max_bb_count);
   for (std::string line; std::getline(bhive_csv_file, line);) {
+    if (file_counter >= max_bb_count) break;
```
ondrasej, line 110: In isolation, this is fine. But we should be careful once we do the JSON converter: do we want to limit the number of blocks or the number of files?

ondrasej (approved):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: We still want it to cover the number of blocks as this is mostly intended to make quick experiments on large datasets easier (without having to cut the CSV up), and the time taken to annotate should be roughly linear with the number of blocks. I'll make sure to rebase the JSON converter on top of this.

PR #36, opened 2024-01-27
Bump LLVM version

boomanaiden154:
This patch bumps the LLVM version to a recent ToT commit and patches the Exegesis annotator so that everything compiles and works correctly.

Mostly wanting to get this in as I remember debugging weird failures due to things compiling despite a constructor changing within upstream Exegesis.


ondrasej:
Thanks for working on this!

PR #38, opened 2024-01-27
Add JSON output option to BHive converter

boomanaiden154:
This patch adds a JSON output option to the BHive converter. This makes it significantly easier to implement other scripts down the line that ingest this data. This also cuts down the number of inodes that a large data set will use by a significant amount, which can be a problem on some file systems.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -42,19 +42,16 @@ constexpr std::string_view kMemNamePrefix = "MEM";
 
 ABSL_FLAG(std::string, bhive_csv, "", "Filename of the input BHive CSV file");
 ABSL_FLAG(
-    std::string, output_dir, "",
+    std::string, asm_output_dir, "",
     "Directory containing output files that can be executed by llvm-exegesis");
 ABSL_FLAG(std::string, json_output_dir, "",
           "Directory containing JSON output files");
 ABSL_FLAG(
     unsigned, json_split_count, std::numeric_limits<unsigned>::max(),
     "The number of annotated basic blocks to include in a single JSON file");
-ABSL_FLAG(std::vector<std::string>, output_types,
-          std::vector<std::string>({"asm"}),
-          "A comma separated list of output types to generate");
 
-bool write_json_file(llvm::json::Array to_write, size_t json_file_number,
-                     std::string json_output_dir) {
+bool WriteJsonFile(llvm::json::Array to_write, size_t json_file_number,
+                   std::string json_output_dir) {
   llvm::Twine json_output_file_path = llvm::Twine(json_output_dir)
                                           .concat("/")
                                           .concat(llvm::Twine(json_file_number))
@@ -83,27 +80,7 @@ int main(int argc, char* argv[]) {
   }
 
   const std::string json_output_dir = absl::GetFlag(FLAGS_json_output_dir);
-  const std::string output_dir = absl::GetFlag(FLAGS_output_dir);
-  const std::vector<std::string> output_types =
-      absl::GetFlag(FLAGS_output_types);
-  bool json_output_enabled = false;
-  bool asm_output_enabled = false;
-  for (const std::string& output_type : output_types) {
-    if (output_type == "json")
-      json_output_enabled = true;
-    else if (output_type == "asm")
-      asm_output_enabled = true;
-  }
-
-  if (json_output_enabled && json_output_dir.empty()) {
-    std::cerr << "Error: --json_output_dir is required when the json output "
-                 "type is requested\n";
-    return 1;
-  } else if (asm_output_enabled && output_dir.empty()) {
-    std::cerr << "Error: --output_dir is required when the asm output type is "
-                 "requested\n";
-    return 1;
-  }
+  const std::string asm_output_dir = absl::GetFlag(FLAGS_asm_output_dir);
 
   std::string initial_reg_val_str =
       gematria::ConvertHexToString(kInitialRegVal);
@@ -187,9 +164,9 @@ int main(int argc, char* argv[]) {
       continue;
     }
 
-    if (asm_output_enabled) {
+    if (!asm_output_dir.empty()) {
       // Create output file path.
-      llvm::Twine output_file_path = llvm::Twine(output_dir)
+      llvm::Twine output_file_path = llvm::Twine(asm_output_dir)
                                          .concat("/")
                                          .concat(llvm::Twine(file_counter))
                                          .concat(".test");
@@ -221,15 +198,14 @@ int main(int argc, char* argv[]) {
       }
     }
 
-    if (json_output_enabled) {
+    if (!json_output_dir.empty()) {
       llvm::json::Object current_snippet;
 
       if (addrs->accessed_blocks.size() > 0) {
         llvm::json::Array memory_definitions;
         llvm::json::Object current_memory_definition;
         current_memory_definition["Name"] = llvm::json::Value(kMemNamePrefix);
-        current_memory_definition["Size"] =
-            llvm::json::Value(addrs->block_size);
+        current_memory_definition["Size"] = addrs->block_size;
         current_memory_definition["Value"] = llvm::json::Value(kInitialMemVal);
         memory_definitions.push_back(std::move(current_memory_definition));
         current_snippet["MemoryDefinitions"] =
@@ -239,7 +215,7 @@ int main(int argc, char* argv[]) {
         for (const uintptr_t addr : addrs->accessed_blocks) {
           llvm::json::Object current_memory_mapping;
           current_memory_mapping["Value"] = llvm::json::Value(kMemNamePrefix);
-          current_memory_mapping["Address"] = llvm::json::Value(addr);
+          current_memory_mapping["Address"] = addr;
           memory_mappings.push_back(std::move(current_memory_mapping));
         }
         current_snippet["MemoryMappings"] =
@@ -249,15 +225,14 @@ int main(int argc, char* argv[]) {
         current_snippet["MemoryMappings"] = llvm::json::Array();
       }
 
-      std::string hex_string = {hex.begin(), hex.end()};
-      current_snippet["Hex"] = llvm::json::Value(hex_string);
+      current_snippet["Hex"] = std::string(hex);
 
       processed_snippets.push_back(
           llvm::json::Value(std::move(current_snippet)));
 
       if (file_counter % json_split_count == 0) {
         size_t json_file_number = file_counter / json_split_count;
-        bool write_successfully = write_json_file(
+        bool write_successfully = WriteJsonFile(
             std::move(processed_snippets), json_file_number, json_output_dir);
         if (!write_successfully) return 4;
         processed_snippets.clear();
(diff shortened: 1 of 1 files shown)


gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -44,6 +48,34 @@ ABSL_FLAG(std::string, bhive_csv, "", "Filename of the input BHive CSV file");
 ABSL_FLAG(
     std::string, output_dir, "",
     "Directory containing output files that can be executed by llvm-exegesis");
+ABSL_FLAG(std::string, json_output_dir, "",
+          "Directory containing JSON output files");
+ABSL_FLAG(
+    unsigned, json_split_count, std::numeric_limits<unsigned>::max(),
+    "The number of annotated basic blocks to include in a single JSON file");
+ABSL_FLAG(std::vector<std::string>, output_types,
```
mtrofin, line 56: isn't there some `enum` based flag that'd self-validate instead?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -54,9 +86,26 @@ int main(int argc, char* argv[]) {
...
 
+  const std::string json_output_dir = absl::GetFlag(FLAGS_json_output_dir);
   const std::string output_dir = absl::GetFlag(FLAGS_output_dir);
-  if (output_dir.empty()) {
-    std::cerr << "Error: --output_dir is required\n";
+  const std::vector<std::string> output_types =
+      absl::GetFlag(FLAGS_output_types);
+  bool json_output_enabled = false;
+  bool asm_output_enabled = false;
+  for (const std::string& output_type : output_types) {
```
mtrofin, line 95: why do we need to support multiple simultaneous outputs?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -54,9 +86,26 @@ int main(int argc, char* argv[]) {
...
+      json_output_enabled = true;
+    else if (output_type == "asm")
+      asm_output_enabled = true;
+  }
+
+  if (json_output_enabled && json_output_dir.empty()) {
+    std::cerr << "Error: --json_output_dir is required when the json output "
+                 "type is requested\n";
+    return 1;
+  } else if (asm_output_enabled && output_dir.empty()) {
```
mtrofin, line 106: ok, so you're only supporting one output. so might as well type the flag as std::string (or an enum and self-validate)

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -44,6 +48,34 @@ ABSL_FLAG(std::string, bhive_csv, "", "Filename of the input BHive CSV file");
...
+ABSL_FLAG(std::string, json_output_dir, "",
+          "Directory containing JSON output files");
+ABSL_FLAG(
+    unsigned, json_split_count, std::numeric_limits<unsigned>::max(),
+    "The number of annotated basic blocks to include in a single JSON file");
+ABSL_FLAG(std::vector<std::string>, output_types,
+          std::vector<std::string>({"asm"}),
+          "A comma separated list of output types to generate");
+
+bool write_json_file(llvm::json::Array to_write, size_t json_file_number,
```
ondrasej, line 60: The Google style guide uses `BigCamelCase` for function and method names.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -54,9 +86,26 @@ int main(int argc, char* argv[]) {
...
+  const std::vector<std::string> output_types =
+      absl::GetFlag(FLAGS_output_types);
+  bool json_output_enabled = false;
+  bool asm_output_enabled = false;
+  for (const std::string& output_type : output_types) {
+    if (output_type == "json")
+      json_output_enabled = true;
+    else if (output_type == "asm")
+      asm_output_enabled = true;
+  }
```
ondrasej, line 100: Simpler alternative if we do want to allow both types of output at the same time: just have two `bool` flags "FLAGS_json_output" and "FLAGS_asm_output".

Or even simpler: since we have different flags for output dirs, all we really need is:
- `FLAGS_json_output_dir` and `FLAGS_asm_output_dir`, and enable the output when the flag is non-empty. The output types can then be completely removed, and we wouldn't even need this check.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -139,38 +190,90 @@ int main(int argc, char* argv[]) {
...
-    // Append disassembled instructions.
-    for (const auto& instr : proto->machine_instructions()) {
-      output_file << instr.assembly() << "\n";
+    if (json_output_enabled) {
+      llvm::json::Object current_snippet;
+
+      if (addrs->accessed_blocks.size() > 0) {
+        llvm::json::Array memory_definitions;
+        llvm::json::Object current_memory_definition;
+        current_memory_definition["Name"] = llvm::json::Value(kMemNamePrefix);
```
ondrasej, line 233: IIUC the code right, `llvm::json::Value` has implicit conversions, and it should be possible to write just
```
current_memory_definition["Name"] = kMemNamePrefix;
```
and so on for the others.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -139,38 +190,90 @@ int main(int argc, char* argv[]) {
...
+        }
+        current_snippet["MemoryMappings"] =
+            llvm::json::Value(std::move(memory_mappings));
+      } else {
+        current_snippet["MemoryDefinitions"] = llvm::json::Array();
+        current_snippet["MemoryMappings"] = llvm::json::Array();
+      }
+
+      std::string hex_string = {hex.begin(), hex.end()};
+      current_snippet["Hex"] = llvm::json::Value(hex_string);
```
ondrasej, line 256:
```
current_snippet["Hex"] = std::string(hex);
```
doesn't work?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -139,38 +190,90 @@ int main(int argc, char* argv[]) {
...
-}
\ No newline at end of file
+
+  if (json_output_enabled) {
+    size_t json_file_number = file_counter / json_split_count;
+    bool write_successfully = write_json_file(
+        std::move(processed_snippets), json_file_number, json_output_dir);
+    if (!write_successfully) return 4;
+  }
+}
```
ondrasej, line 279: We're missing `return 0;` at the end. It was missing even before, but since the file is already changed in this PR...

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: All of the enum flags still need to implement AbslParseFlag and AbslUnparseFlag and I wasn't able to get that working with a vector of enums. I've switched to Ondrej's suggestion and now just have flags for the directories and enable the output if the directory is set.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Mostly a convenience thing. It's nice to be able to reduce test cases by hand using the `llvm-exegesis` binary and it would make it easy to compare against a Gematria benchmarking script. This is especially useful given how slow the Exegesis annotator is. We still want JSON output for the benchmarking script and to limit the number of files on disk for large datasets.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: This was meant to be another `if` statement. Sorry for the confusion.

I've switched to just looking at the output directories and have removed this code.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. Switched to just using the output directories.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: `kMemNamePrefix` is a `std::string_view` and it doesn't look like there's an explicit conversion for that one. So we either have to do `= std::string(kMemNamePrefix)` or `= llvm::json::Value(kMemNamePrefix)`.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Fixed. Not sure how I ended up with that weirdness.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -38,8 +42,33 @@ constexpr std::string_view kMemNamePrefix = "MEM";
...
+ABSL_FLAG(
+    unsigned, json_split_count, std::numeric_limits<unsigned>::max(),
+    "The number of annotated basic blocks to include in a single JSON file");
+
+bool WriteJsonFile(llvm::json::Array to_write, size_t json_file_number,
+                   std::string json_output_dir) {
+  llvm::Twine json_output_file_path = llvm::Twine(json_output_dir)
+                                          .concat("/")
+                                          .concat(llvm::Twine(json_file_number))
+                                          .concat(".json");
```
ondrasej, line 58: I know this comes from the previous code, but still:

Ideally, we'd use `llvm::path::append()` here to automatically correctly handle separators, or at least we should be using `llvm::path::get_separator()` to get the right path separator.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -98,6 +124,8 @@ int main(int argc, char* argv[]) {
   gematria::BHiveImporter bhive_importer(&canonicalizer);
 
   std::ifstream bhive_csv_file(bhive_filename);
+  llvm::json::Array processed_snippets;
+  const int json_split_count = absl::GetFlag(FLAGS_json_split_count);
```
ondrasej, line 128: We should have a check that `json_split_count > 0`. Otherwise, there might be some divisions by zero.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Switched to using `llvm::path::append()` along with `llvm::SmallString` here.

ondrasej (approved):
One more comment. Feel free to ~submit~ merge once it is resolved.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -47,20 +48,19 @@ ABSL_FLAG(
...
+  llvm::SmallString<40> json_output_file_path(json_output_dir);
+  llvm::sys::path::append(json_output_file_path,
+                          llvm::Twine(json_file_number).concat(".json"));
   std::error_code file_ec;
-  llvm::raw_fd_ostream json_output_file(json_output_file_path.str(), file_ec);
+  llvm::raw_fd_ostream json_output_file(json_output_file_path, file_ec);
 
   if (file_ec) {
-    std::cerr << "Failed to open output file: " << json_output_file_path.str()
+    std::cerr << "Failed to open output file: " << json_output_file_path.c_str()
```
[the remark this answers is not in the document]
boomanaiden154: I don't think `c_str()` should be that much more expensive? It pretty much just returns a pointer to the internal buffer here, which is also what ends up happening when converting to a `std::string`. I've changed it to ` as `static_cast<std::string_view>` just returns a StringRef which doesn't have standard output operators implemented (as they use the LLVM ones). `StringRef::str()` returns a `std::string` and doesn't automatically use the `std::string_view` conversion.

Implementing a `std::string_view` conversion for `llvm::SmallString` would make quite a bit of sense and I'll work on getting something up for that.

Let me know if the approach here needs some adjustment and I can open up a new PR fixing it.

PR #39, opened 2024-01-27
Change constants in BHive conversion script

boomanaiden154:
This patch adjusts the constants in the BHive conversion script from the default values added in with the script to the values used within the original BHive paper. These constants were experimentally determined within the BHive paper to be high enough to not underflow/access addresses in the first page yet low enough to not access any addresses above the virtual address space ceiling. Anecdotally, these constants have also seemed to work better than the original ones. In addition, this patch prefixes the memory value with zeroes so that llvm-exegesis is able to assume the correct bit width.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -29,6 +29,10 @@
 #include "gematria/llvm/llvm_architecture_support.h"
 #include "gematria/utils/string.h"
 
+// Use the constants from the BHive paper for setting initial register and
+// memory values. These constants are set to a high enough value to avoid
+// underflow and accesses within the first page, but low enough to avoid
+// exceeding the virtual address space ceiling in most cases.
 constexpr uint64_t kInitialRegVal = 0x12345600;
 constexpr uint64_t kInitialMemVal = 0x12345600;
 constexpr unsigned kInitialMemValBitWidth = 64;
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -29,8 +29,9 @@
 #include "gematria/llvm/llvm_architecture_support.h"
 #include "gematria/utils/string.h"
 
-constexpr uint64_t kInitialRegVal = 0x10000;
-constexpr uint64_t kInitialMemVal = 0x7FFFFFFF;
+constexpr uint64_t kInitialRegVal = 0x12345600;
+constexpr uint64_t kInitialMemVal = 0x12345600;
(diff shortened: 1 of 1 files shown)


mtrofin:
> the values used within the original BHive paper. Anecdotally, these constants have also seemed to work better than the original ones.

which are the original counter values, the ones in the original paper or the ones before this change?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -29,8 +29,9 @@
 #include "gematria/llvm/llvm_architecture_support.h"
 #include "gematria/utils/string.h"
 
-constexpr uint64_t kInitialRegVal = 0x10000;
-constexpr uint64_t kInitialMemVal = 0x7FFFFFFF;
+constexpr uint64_t kInitialRegVal = 0x12345600;
```
mtrofin, line 32: add a comment about where the numbers came from, because they are quite odd otherwise

do we understand why they are better, exactly?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -64,6 +65,14 @@ int main(int argc, char* argv[]) {
       gematria::ConvertHexToString(kInitialRegVal);
   std::string initial_mem_val_str =
       gematria::ConvertHexToString(kInitialMemVal);
+  // Prefix the string with zeroes as llvm-exegesis assumes the bit width
```
mtrofin, line 68: this is unrelated?

boomanaiden154:
> which are the original counter values, the ones in the original paper or the ones before this change?

The ones before this change. We're switching to the values in the BHive paper here.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Slightly unrelated, but necessary to make this patch work properly. We're assuming the memory value has a width of 64 bits, but only setting the first x bits. Exegesis determines the width of the value based on how many characters are in the string, so we need to explicitly print out the leading zeros.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Comment added along with an explanation for the second question. Essentially they are high enough to avoid accesses in the first page/underflow, but low enough where most accesses are below the virtual address space ceiling. They're not perfect, but anecdotally they work reasonably well.

mtrofin:
> > which are the original counter values, the ones in the original paper or the ones before this change?
> 
> The ones before this change. We're switching to the values in the BHive paper here.

Ack, if you can update the patch description

mtrofin (approved):

PR #40, opened 2024-01-27
Address PR feedback in convert_bhive_to_llvm_exegesis_input

boomanaiden154:
This patch moves `file_counter` out of the global scope to where it is relevant in `main`, as requested in the original review.


ondrasej:
Thanks for cleaning this up!

PR #41, opened 2024-01-28
Add support for exegesis annotator in BHive conversion script

boomanaiden154:
This patch adds support for using the exegesis annotator in the BHive conversion script.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -25,9 +25,11 @@
 #include "absl/flags/parse.h"
 #include "gematria/datasets/bhive_importer.h"
 #include "gematria/datasets/find_accessed_addrs.h"
+#include "gematria/datasets/find_accessed_addrs_exegesis.h"
 #include "gematria/llvm/canonicalizer.h"
 #include "gematria/llvm/llvm_architecture_support.h"
 #include "gematria/utils/string.h"
+#include "llvm/tools/llvm-exegesis/lib/TargetSelect.h"
 
 constexpr uint64_t kInitialRegVal = 0x10000;
 constexpr uint64_t kInitialMemVal = 0x7FFFFFFF;
@@ -40,10 +42,60 @@ namespace {
 unsigned int file_counter = 0;
 }
 
+enum class annotator_type { exegesis, fast };
+
+bool AbslParseFlag(absl::string_view text, annotator_type* type,
+                   std::string* error) {
+  if (text == "exegesis") {
+    *type = annotator_type::exegesis;
+    return true;
+  } else if (text == "fast") {
+    *type = annotator_type::fast;
+    return true;
+  }
+
+  *error = "unknown annotator type";
+  return false;
+}
+
+std::string AbslUnparseFlag(annotator_type type) {
+  switch (type) {
+    case annotator_type::exegesis:
+      return "exegesis";
+    case annotator_type::fast:
+      return "fast";
+  }
+}
+
 ABSL_FLAG(std::string, bhive_csv, "", "Filename of the input BHive CSV file");
 ABSL_FLAG(
     std::string, output_dir, "",
     "Directory containing output files that can be executed by llvm-exegesis");
+ABSL_FLAG(annotator_type, annotator_implementation, annotator_type::fast,
+          "The annotator implementation to use.");
+
+std::optional<gematria::AccessedAddrs> GetAccessedAddrs(
+    absl::Span<const uint8_t> basic_block,
+    gematria::ExegesisAnnotator* exegesis_annotator) {
+  const annotator_type annotator_implementation =
+      absl::GetFlag(FLAGS_annotator_implementation);
+  if (annotator_implementation == annotator_type::fast) {
+    // This will only get the first segfault address.
+    auto addrs = gematria::FindAccessedAddrs(basic_block);
+
+    if (!addrs.ok()) return std::nullopt;
+
+    return *addrs;
+  } else if (annotator_implementation == annotator_type::exegesis) {
+    auto addrs = exegesis_annotator->findAccessedAddrs(
+        llvm::ArrayRef(basic_block.begin(), basic_block.end()));
+
+    if (!addrs) return std::nullopt;
+
+    return *addrs;
+  }
+  return std::nullopt;
+}
 
 int main(int argc, char* argv[]) {
   absl::ParseCommandLine(argc, argv);
@@ -59,6 +111,9 @@ int main(int argc, char* argv[]) {
     std::cerr << "Error: --output_dir is required\n";
     return 1;
   }
+
+  const annotator_type annotator_implementation =
+      absl::GetFlag(FLAGS_annotator_implementation);
 
   std::string initial_reg_val_str =
       gematria::ConvertHexToString(kInitialRegVal);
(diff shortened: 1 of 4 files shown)


gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -40,10 +42,60 @@ namespace {
 unsigned int file_counter = 0;
 }
 
+enum class annotator_type { exegesis, fast };
```
ondrasej, line 45: Nit: This should follow the Google naming conventions with
`AnnotatorType`, `kExegesis`, and `kFast`.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -40,10 +42,60 @@ namespace {
...
+}
+
+std::string AbslUnparseFlag(annotator_type type) {
+  switch (type) {
+    case annotator_type::exegesis:
+      return "exegesis";
+    case annotator_type::fast:
+      return "fast";
+  }
+}
```
ondrasej, line 68: Optional: I'd define those as something like
```
constexpr std::pair<AnnotatorType, std::string_view> kAnnotatorTypeNames = {
    {AnnotatorType::kExegesis, "exegesis"},
    {AnnotatorType::kFast, "fast"}
};
```
then parsing and unparsing can both be done with a single for loop, and if we ever add more values, only one place needs to be changed.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -40,10 +42,60 @@ namespace {
...
     "Directory containing output files that can be executed by llvm-exegesis");
+ABSL_FLAG(annotator_type, annotator_implementation, annotator_type::fast,
+          "The annotator implementation to use.");
+
+std::optional<gematria::AccessedAddrs> GetAccessedAddrs(
+    absl::Span<const uint8_t> basic_block,
+    gematria::ExegesisAnnotator* exegesis_annotator) {
+  const annotator_type annotator_implementation =
+      absl::GetFlag(FLAGS_annotator_implementation);
+  if (annotator_implementation == annotator_type::fast) {
```
ondrasej, line 82: Maybe a switch statement? (though for two values, this is fine)

ondrasej (changes requested):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Is that supposed to be a `constexpr std::array<std::pair<AnnotatorType, std::string_view>,2>`? We don't have `constexpr std::vector` as we aren't using C++20. I've switched to that and it's definitely a lot cleaner.

I used a `__builtin_unreachable` for the `AbslUnparseFlag` as clang complains about not having a return at the end of the function even though that should never be reached. Not sure what the typical practice for this is in Google C++.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -36,10 +38,65 @@ constexpr std::string_view kMemDefPrefix = "# LLVM-EXEGESIS-MEM-DEF ";
...
+
+std::optional<gematria::AccessedAddrs> GetAccessedAddrs(
+    absl::Span<const uint8_t> basic_block,
+    gematria::ExegesisAnnotator* exegesis_annotator) {
+  const AnnotatorType annotator_implementation =
+      absl::GetFlag(FLAGS_annotator_implementation);
+  switch (annotator_implementation) {
+    case AnnotatorType::kFast: {
+      // This will only get the first segfault address.
+      auto addrs_fast = gematria::FindAccessedAddrs(basic_block);
```
mtrofin, line 83: See gematria/llvm/llvm_to_absl.h, maybe add these backward converters there? same for line 93.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -97,6 +157,25 @@ int main(int argc, char* argv[]) {
   gematria::X86Canonicalizer canonicalizer(&llvm_support->target_machine());
   gematria::BHiveImporter bhive_importer(&canonicalizer);
 
+  llvm::exegesis::InitializeX86ExegesisTarget();
+
+  auto llvm_state_or_error = llvm::exegesis::LLVMState::Create("", "native");
+  if (!llvm_state_or_error) {
+    std::cerr << "Failed to create LLVMState\n";
```
mtrofin, line 164: we don't have absl::log?

mtrofin (commented):
I assume this file is OK with abseil deps @ondrasej ?

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: We could, but the rest of the file uses standard library output, so I'm staying consistent with that for now. Not sure what would be preferred though. I think within Google, everything would use `absl::log`, and LLVM has its own output facilities, so this is sort of in no man's land.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: `gematria::FindAccessedAddrs` already returns a `absl::StatusOr`. We could switch line 93, but I figured since we're just checking the error within the case and returning the actual value, it didn't make a lot of sense.

mtrofin: No, I mean add a  `absl::StatusOr` -> `std::optional` converter and then reuse it here and below

boomanaiden154: I've added a `absl::StatusOr` to `std::optional` converter in `llvm_to_absl.h`. I've used `llvm::expectedToOptional` to convert the expected to an optional value.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
ondrasej: Oops, sorry, I forgot the array brackets. What I meant was a plain old array
```
constexpr std::pair<AnnotatorType, std::string_view> kAnnotatorTypeNames[] = {
    {AnnotatorType::kExegesis, "exegesis"},
    {AnnotatorType::kFast, "fast"}
};
```
You can get STL-compatible iterators for it with `std::begin()` and `std::end()`, so all the `std::find()` would work. Nice thing about the plain old array is that it doesn't need to have the size specified.

But `std::array<...>` is fine too.

ondrasej:
> I assume this file is OK with abseil deps @ondrasej ?

I think this one is fine. We're using LLVM from this one, but we don't need tflite, so Bazel build should not be a problem.

It would be nice to get rid of Abseil eventually, but it's not necessary in this PR.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Ah, makes sense. I'll switch to a normal array to avoid the extra parameter in the `std::array` template. Any preference on a for loop vs `std::find`?

ondrasej: No preference, whatever is simpler/shorter :)

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
ondrasej: Since it is an error, should we convert to `llvm::Expected`?

boomanaiden154: I originally structured it this way given that the handler had additional useful information to report (in the calling function) that didn't make sense to send through here. Errors were originally reported by printing them out, but with the refactoring to the converters, that went away.

I guess we should probably just be returning a `absl::StatusOr` and converting the expected into that. Not sure why I didn't think of that in the first place.

boomanaiden154: Fixed. Sorry for the back and forth here/misunderstanding on my end.

ondrasej (approved):
I don't have any more comments.

PR #42, opened 2024-02-02
Add test target for llvm-cm

boomanaiden154:
This patch sets up a ninja target for testing llvm-cm, akin to the tool specific test targets in the monorepo. This allows for actually running the tests. Documentation has been added to the README.

### DIFF
--- a/llvm_cm/test/X86/bb-frequency.s
+++ b/llvm_cm/test/X86/bb-frequency.s
@@ -1,5 +1,4 @@
 ## LLVM-CM frequency weighting test.
-# REQUIRES: x86
 # RUN: split-file %s %t
 # RUN: llvm-mc -o %t.o --filetype=obj -triple=x86_64-unknown-linux-gnu %t/bb-frequency-test.s
 # RUN: llvm-cm %t.o --csv=%t/bb-frequency.csv -granite_model=%S/Inputs/gb-token-mit-2022_12_02.tflite -evaluator=granite| FileCheck %t/bb-frequency-test.s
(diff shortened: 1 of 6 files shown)


boomanaiden154:
@JestrTulip

llvm_cm/test/X86/bb-frequency.s
```diff
@@ -1,5 +1,4 @@
 ## LLVM-CM frequency weighting test.
-# REQUIRES: x86
```
mtrofin, line 2: why

llvm_cm/test/lit.cfg
```diff
@@ -0,0 +1,19 @@
...
+
+config.name = "llvm-cm"
+config.test_format = lit.formats.ShTest(execute_external=False)
+
+config.suffixes = [".s", ".test"]
+
+config.test_source_root = os.path.dirname(__file__)
+config.test_exec_root = config.obj_root
+
+llvm_config.use_default_substitutions()
```
mtrofin, line 15: is there a way to inherit substitutions the llvm test cfg.in makes?

boomanaiden154: Not from what I can tell. It just uses the `lit.cfg` at the root of the test directory along with `lit.site.cfg` for subdirectories, and since we're not in `llvm/test`, we don't get to take advantage of that.

llvm_cm/test/X86/bb-frequency.s
boomanaiden154: Was testing things and removed it, figuring that this isn't likely to get run on anything other than X86 currently. I've switched it to use `x86-registered-target` as I believe that's all it should need. (Unless the tflite models don't work on something like AArch64?).

mtrofin: They would on that, as in afaik the interpreter can be built there, but probably won't on other hosts for which llvm can be compiled

boomanaiden154: Ack. Switched to only running on `x86-64` using that native lit requirements.

mtrofin (approved):

PR #43, opened 2024-02-02
Fix llvm-cm after switch to BBRanges

boomanaiden154:
A couple additional cases were recently added into the BB Address Map section in https://github.com/llvm/llvm-project/pull/74128. This broke llvm-cm as a couple method names changed. This patch provides the quick fix to get everything working.

Eventually llvm-cm will need some work to support multiple BBRanges within a single function.


boomanaiden154:
@JestrTulip

boomanaiden154:
In addition to supporting BBRanges, `llvm-cm` also needs to use `PGOAnalysisMap` as `mbb-profile-dump` has been removed from upstream.

mtrofin (approved):

PR #47, opened 2024-02-23
Add progress reporting flag to annotator script

boomanaiden154:
This patch adds a progress reporting flag to the annotator script. This can be quite useful when using a slower implementation of FindAccessedAddrs like FindAccessedAddrsExegesis to get a gauge on what sort of progress is being made and to show that the application isn't stuck somewhere when processing a large number of blocks.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -88,9 +88,8 @@ ABSL_FLAG(
     "The number of annotated basic blocks to include in a single JSON file");
 ABSL_FLAG(unsigned, max_bb_count, std::numeric_limits<unsigned>::max(),
           "The maximum number of basic blocks to process");
-ABSL_FLAG(
-    bool, report_progress, false,
-    "Whether or not to report progress after annotating each basic block.");
+ABSL_FLAG(unsigned, report_progress_every, std::numeric_limits<unsigned>::max(),
+          "The number of blocks after which to report progress.");
 
 absl::StatusOr<gematria::AccessedAddrs> GetAccessedAddrs(
     absl::Span<const uint8_t> basic_block,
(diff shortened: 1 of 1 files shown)


gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -332,6 +336,9 @@ int main(int argc, char* argv[]) {
       }
     }
 
+    if (report_progress)
+      std::cout << "Finished annotating block #" << file_counter << ".\n";
```
ondrasej, line 340: I'd put all service messages (errors, warnings, but even progress) to stderr. stdout should be reserved for actual requested output of the program (we have none here, because we always write to a file).

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -332,6 +336,9 @@ int main(int argc, char* argv[]) {
       }
     }
 
+    if (report_progress)
```
ondrasej, line 339: I think reporting each 100th file would be sufficient and it would create less noise in the output. Of course, depending on the speed - but if we process 1000s of blocks per second, this might be overwhelming.

Perhaps, we could even turn the current flag into a flag saying after how many blocks to report progress (and turn it off by setting it to a high value, e.g. maxint).

ondrasej (changes requested):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. Updated this to rename the flag `report_progress_every` that takes an unsigned integer value and reports status after every n blocks.

The exegesis annotator definitely doesn't run that fast (maybe a couple blocks a second from my testing), so it's not an issue there, but using Owen's annotator implementation would definitely end up dumping a massive amount of text very quickly to the console.

ondrasej (approved):

PR #49, opened 2024-02-24
Fix off by one error in annotator JSON output

boomanaiden154:
This patch fixes an off-by-one error in the annotator JSON output. Essentially, the very first block would automatically trigger a write to the file since the file count was zero, and zero modulo anything is still zero, which was the condition for writing to a file. This becomes a problem if there are fewer blocks than blocks_per_json_file as after processing all the blocks, the final chunk will be written, but it will overwrite the first chunk in this case, leaving blocks out. This patch fixes that behavior by simpling computing the module with file_counter +1, which has the same effect but avoids this behavior.


boomanaiden154:
This (and other issues) I think definitely highlight the need for lit tests. I'd really like to get the infrastructure for that up and running, but having to get it to work with Bazel doesn't seem particularly fun.

ondrasej (approved):

PR #51, opened 2024-02-29
Only add definitions for used registers

boomanaiden154:
This patch makes the BHive to Exegesis conversion script only include definitions for registers needed by the snippet rather than a random set. This makes the outputs cleaner and will help keep things clean later on when setting a loop register (and also make information available to determine if/where there are conflicts).


mtrofin:
should it have a test?

boomanaiden154:
Existing tests have been updated (which should be enough coverage here) now that testing infrastructure is  in place.

boomanaiden154:
I accidentally merged in #52 into this branch (tried to do stacked PRs and forgot how the DAG was setup while being overly zealous with the merge button). I can probably fix this if it's really necessary but the only changes from that PR (which was approved) are `json.test`, and lines 305-314.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -240,22 +215,47 @@ int main(int argc, char* argv[]) {
...
+      std::cerr << "Failed to disassemble block '" << hex << "\n";
       continue;
     }
 
+    // Get used registers
+    std::map<unsigned, bool> used_registers;
+    for (const gematria::DisassembledInstruction& instruction : *instructions) {
+      for (unsigned operand_index = 0;
+           operand_index < instruction.mc_inst.getNumOperands();
+           ++operand_index) {
```
ondrasej, line 235: With this, we enumerate inputs as well as outputs. Ideally, we should go from `instr_desc.getNumDefs()` to `mc_inst.getNumOperands()` to capture only the inputs.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -240,22 +215,47 @@ int main(int argc, char* argv[]) {
...
+        if (instruction.mc_inst.getOperand(operand_index).isReg()) {
+          unsigned register_number =
+              instruction.mc_inst.getOperand(operand_index).getReg();
+          if (register_number == 0) continue;
+          unsigned super_register_number = 0;
+          for (MCPhysReg testing :
+               reg_info.superregs_inclusive(register_number)) {
+            super_register_number = testing;
+          }
```
ondrasej, line 244: This stops at the _last_ super-register. Is there a guarantee that it is also the largest super-register? If not, we should either recurse (whenever we find a proper super-register, step up and look for its own super-registers), or filter by register class and take only the 64-bit superreg, if any.

Though the general idea is probably safer and more general.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -240,22 +215,47 @@ int main(int argc, char* argv[]) {
...
+          used_registers[super_register_number] = true;
+        }
```
ondrasej, line 246: We also need to look at implicit input registers, via `MCInstrDesc::implicit_uses()`.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -240,22 +215,47 @@ int main(int argc, char* argv[]) {
...
+      }
+    }
```
ondrasej, line 248: Thinking about it a bit more, this should go to a library function where we can also add proper unit tests.

ondrasej (changes requested):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: From everything that I can tell, yes. It's not explicitly enumerated in the documentation for the iterator, but the iterator internally starts at the given register and then goes up one level at a time (from what I can tell). Would probably be good to open a patch on the LLVM side to document this.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. Split into #71, where the actual functionality is implemented.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. Switched over to iterating through the inputs of the register only (and keeping track of what registers get defined in the snippet so we can omit them from the usage list).

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. Functionality for that has been added in #71.

boomanaiden154:
This is now part of a PR stack, depending on #71.

ondrasej:
> This is now part of a PR stack, depending on #71.

SG, let's get #71 merged first.

ondrasej:
I'll wait with reviewing until #71 is merged and we have this based just on main.

boomanaiden154:
> I'll wait with reviewing until https://github.com/google/gematria/pull/71 is merged and we have this based just on main.

I've merge main into the branch, so the diff should be clean now.

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,126 @@
...
+
+namespace gematria {
+
+unsigned getSuperRegister(unsigned OriginalRegister,
+                          const MCRegisterInfo &RegisterInfo) {
+  // We should be able to get the super register by getting an iterator from
+  // RegisterInfo and getting the back, but this always returns a value of
+  // zero.
+  // TODO(boomanaiden154): Investigate why this is ocurring on the LLVM side
+  // and fix it.
```
ondrasej, line 35: No action needed, just discussion.

If I understand the comment in [MCRegisterInfo.h](https://github.com/llvm/llvm-project/blob/00c622e596f918d9d83674b58097c8982ae1af95/llvm/include/llvm/MC/MCRegisterInfo.h#L109) correctly, the superregs are stored in a zero-terminated array, so what we see here is probably the terminator (and we need to do something like back() - 1)?

Either way, this seems to be safe, and unless it is documented somewhere that the last value is the largest one, I'd stick with the slower but safe code.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -251,23 +227,31 @@ int main(int argc, char* argv[]) {
...
-    // Check for errors.
-    if (!proto.ok()) {
-      std::cerr << "Failed to disassemble block '" << hex
-                << "': " << proto.status() << "\n";
+    if (!instructions) {
+      std::cerr << "Failed to disassemble block '" << hex << "\n";
       continue;
     }
 
+    // Get used registers
```
ondrasej, line 241: Very nit: Add a period at the end of the line.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -251,23 +227,31 @@ int main(int argc, char* argv[]) {
...
+            llvm_support->mc_disassembler(), llvm_support->mc_instr_info(),
+            llvm_support->mc_register_info(), llvm_support->mc_subtarget_info(),
+            *inst_printer, 0, *bytes);
 
-    // Check for errors.
-    if (!proto.ok()) {
-      std::cerr << "Failed to disassemble block '" << hex
-                << "': " << proto.status() << "\n";
+    if (!instructions) {
+      std::cerr << "Failed to disassemble block '" << hex << "\n";
```
ondrasej, line 237: The closing apostrophe is missing (we should have `"'\n"` rather than `"\n"` at the end).

ondrasej (approved):
Looks good. I've added a few minor comments, go ahead with the merge once they are resolved.

gematria/datasets/basic_block_utils.cc
boomanaiden154: Ah, that would make sense. Sounds good keeping the safe code for now. I guess if it shows up hot in a profile, we can always adjust, but I don't really see that happening.

PR #53, opened 2024-03-01
Add lit tests for convert_bhive_to_llvm_exegesis_input

boomanaiden154:
This patch adds the necessary infrastructure for creating lit tests for `convert_bhive_to_llvm_exegesis_input`. This patch also adds in a single unit test to demonstrate the functionality.


gematria/datasets/BUILD
```diff
@@ -69,6 +74,19 @@ cc_binary(
     ],
 )
 
+gematria_py_test(
```
mtrofin, line 77: IIUC this makes all the tests appear as one bazel test target - I imagine individual sub-tests still appear as such?

boomanaiden154: Yes. I don't think there's an easy way to get them to show up as individual test targets (but GTest binaries also show up as a single test target from what I understand).

Everything shows up normally in the log output (just like a classical lit invocation), so by default it won't show the individual test cases (although we could pass the flag to do that), but will show explicitly what failed and why.

PR #65, opened 2024-03-08
Add Exegesis Benchmarking tool

### LINKED ISSUE #57: Implement benchmarking script
In order to construct large-scale BB datasets, we need a script that can perform these benchmarking runs, taking in annotated basic blocks from the annotation script (most likely in JSON), and then returning them with throughput information.

boomanaiden154:
Needs some work. Currently just opening this as a placeholder.

Fixes #57.


gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+      MemVal.Index = 0;  // Update this to support multiple definitions
+      MemVal.SizeBytes = *MemoryDefinitionSize;
+
+      BenchCode.Key.MemoryValues[MemoryDefinitionName->str()] = MemVal;
+    }
+
+    const json::Array *MemoryMappings =
+        TestValue.getAsObject()->getArray("MemoryMappings");
+
+    if (!MemoryMappings) ExitWithError("Malformed memory mapping");
```
ondrasej, line 192: I'd put more specific error messages, to make debugging easier. E.g. "Memory mappings are missing.".

That said, this might not be a problem - if the block doesn't access memory, it should be OK to skip this section.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+
+    for (const auto &MemoryMappingValue : *MemoryMappings) {
+      const json::Object *MemoryMappingObject =
+          MemoryMappingValue.getAsObject();
+
+      if (!MemoryMappingObject) ExitWithError("Malformed memory mapping");
```
ondrasej, line 198: Same comment on providing more details in the error messages (e.g. "Memory mapping #i is not a JSON object." or something like that.

In the ideal case, we'd have line numbers in the JSON file, but by now they're probably long lost.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+            Benchmark::RepetitionModeE::MiddleHalfLoop);
+
+    Benchmark Result = std::move(AllResults[0]);
+
+    ResultAgg->AggregateResults(Result,
+                                ArrayRef<Benchmark>(AllResults).drop_front());
+
+    dbgs() << Result.Measurements[0].PerSnippetValue << "\n";
+
+    dbgs() << *HexValue << "\n";
```
ondrasej, line 250: The outputs (the throughput numbers) should go either to a file, or to stdout (to make it easy to filter them out of any possible debug messages that go to stderr).

ondrasej (changes requested):
Overall, this looks good, but:
1. we should have a good story for the outputs. From the Gematria point of view, it would be best if we could write the results into a tfrecord file along with the basic block, and the easiest way to do that would be to wrap this with pybind11 and call it from Python.
2. We should also have a test (if at all possible).

gematria/datasets/BUILD
```diff
@@ -69,6 +69,20 @@ cc_binary(
...
+    name = "exegesis-benchmark",
+    srcs = ["exegesis-benchmark.cc"],
+    deps = [
+        "//gematria/llvm:disassembler",
+        "//gematria/utils:string",
+        "@com_google_absl//absl/flags:flag",
+        "@com_google_absl//absl/flags:parse",
+        "@llvm-project//llvm:Support",
+        "@llvm-project//llvm:Exegesis",
+	"@llvm-project//llvm:X86Disassembler",
```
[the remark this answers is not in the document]
boomanaiden154: Done.

WORKSPACE
```diff
@@ -165,9 +165,9 @@ new_git_repository(
 
 # LLVM and its dependencies
 
-LLVM_COMMIT = "ce75cbeda21e3e4e5d0c47ecf44954caf98acfe8"
+LLVM_COMMIT = "31ab2c4f616d686c06e9b573c8f1a4ae7ad2d8c3"
 
-LLVM_SHA256 = "04b70d6c199b32614a7fa2a321f8aebfb100f76a5445f6c34f6467d42e1e341d"
+LLVM_SHA256 = "4d2ed1cd8f1ce8b1a41002dfcac0ff0f12f0734cc5161192deecd5513e873af5"
```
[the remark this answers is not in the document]
boomanaiden154: Yep. Just had this in here for testing. Have since rebased against `main` with the updated LLVM version.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+  if (Err) {
+    ExitOnErr(createFileError(FileName, std::move(Err)));
+  }
+}
+
+template <typename T>
+T ExitOnFileError(const Twine &FileName, Expected<T> &&E) {
+  ExitOnFileError(FileName, E.takeError());
+  return std::move(*E);
+}
```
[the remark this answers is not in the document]
boomanaiden154: Good point. I've just kept `ExitOnErr` and gotten rid of everything else. I decided not to emit file errors as there is only a single input file, so there doesn't need to be any specificity about which input file the error came from as it isn't ambiguous.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+static cl::opt<std::string> AnnotatedBlocksJson(
+    "annotated-blocks-json",
+    cl::desc("Filename of the JSON file containing annotated basic blocks"),
+    cl::init(""));
+
+static ExitOnError ExitOnErr("exegesis-benchmark error: ");
+
+static void ExitWithError(StringRef ErrorMessage) {
+  ExitOnErr(make_error<StringError>(ErrorMessage, inconvertibleErrorCode()));
+}
```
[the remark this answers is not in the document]
boomanaiden154: Removed, just using `ExitOnErr` now.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+
+static void ExitOnFileError(const Twine &FileName, Error Err) {
```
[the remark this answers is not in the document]
boomanaiden154: Removed, just kept `ExitOnErr`.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+    ExitWithError("--annotated_blocks_json is required");
+    return 1;
+  }
+
+  auto JsonMemoryBuffer = ExitOnFileError(
+      AnnotatedBlocksJson,
+      errorOrToExpected(MemoryBuffer::getFile(AnnotatedBlocksJson, true)));
+
+  auto ParsedAnnotatedBlocks =
+      ExitOnErr(json::parse(JsonMemoryBuffer->getBuffer()));
```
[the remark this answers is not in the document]
boomanaiden154: For now I've just kept the error and gotten rid of the file errors due to the input file not being ambiguous.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+
+  for (const auto &TestValue : *ParsedAnnotatedBlocks.getAsArray()) {
+    std::optional<StringRef> HexValue =
+        TestValue.getAsObject()->getString("Hex");
+    if (!HexValue) ExitWithError("Expected basic block to have hex value");
+
+    std::optional<std::vector<uint8_t>> BytesOr =
+        gematria::ParseHexString(HexValue->str());
+
+    if (!BytesOr.has_value()) ExitWithError("Failed to parse hex value");
```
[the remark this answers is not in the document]
boomanaiden154: Same as above.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+                               State);
+
+  if (pfm::pfmInitialize()) ExitWithError("Failed to initialize libpfm");
+
+  for (const auto &TestValue : *ParsedAnnotatedBlocks.getAsArray()) {
+    std::optional<StringRef> HexValue =
+        TestValue.getAsObject()->getString("Hex");
+    if (!HexValue) ExitWithError("Expected basic block to have hex value");
+
+    std::optional<std::vector<uint8_t>> BytesOr =
```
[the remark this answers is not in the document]
boomanaiden154: Done.

gematria/datasets/exegesis-benchmark.cc
```diff
@@ -0,0 +1,256 @@
...
+    const llvm::MCRegisterInfo &MRI = State.getRegInfo();
+
+    for (unsigned I = 0;
+         I < MRI.getRegClass(X86::GR64_NOREX2RegClassID).getNumRegs(); ++I) {
+      RegisterValue RegVal;
+      RegVal.Register =
+          MRI.getRegClass(X86::GR64_NOREX2RegClassID).getRegister(I);
+      RegVal.Value = APInt(64, 0x12345600);
+      BenchCode.Key.RegisterInitialValues.push_back(RegVal);
+    }
```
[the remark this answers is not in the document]
boomanaiden154: Yep. Should be adjusted now. I originally set it up like this as the annotated JSON from `main` won't be compatible with this currently due to some necessary patches still needing to land.

gematria/datasets/exegesis-benchmark.cc
boomanaiden154: Good point. I've added more information on what specifically is missing. These cases theoretically shouldn't happen that often given the JSON is autogenerated though.

gematria/datasets/exegesis-benchmark.cc
boomanaiden154: I've printed them to `STDOUT` now in the BHive format. Eventually we will probably want some explicit output file, but for now I think this is fine.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -0,0 +1,325 @@
...
+
+Expected<BenchmarkCode> parseJSONBlock(
+    const json::Object &BasicBlockJSON, MCInstPrinter &MachinePrinter,
+    const MCDisassembler &MachineDisassembler, const LLVMState &State) {
+  BenchmarkCode BenchCode;
+
+  std::optional<StringRef> HexValue = BasicBlockJSON.getString("Hex");
+  if (!HexValue)
+    return llvm::make_error<StringError>(
+        errc::invalid_argument, "Expected basic block to have hex value");
```
ondrasej, line 54: This is a general comment, on the whole JSON parsing code, not just this line: we should put in as much information as possible, to allow the user quickly locate the part of the file where the problem is.

In this case, basic block index. For register definitions and memory definitions and mappings, there should be also their index, and so on.

ondrasej (changes requested):

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -0,0 +1,325 @@
...
+  SmallVector<Benchmark, 2> AllResults;
+
+  auto RC1 =
+      BenchRunner.getRunnableConfiguration(BenchCode, 5000, 0, *SnipRepetitor);
+  if (!RC1) return RC1.takeError();
+  auto RC2 =
+      BenchRunner.getRunnableConfiguration(BenchCode, 10000, 0, *SnipRepetitor);
+  if (!RC2) return RC2.takeError();
+
+  std::pair<Error, Benchmark> BenchmarkResult1OrErr =
```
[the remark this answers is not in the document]
boomanaiden154: The variables were named `BenchmarkResult1<suffix>` and `BenchmarkResult2<suffix>` since we need two to run the comparison (rather than 10, which is what the capital O makes it look like). I've switched to A/B to make things more clear.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -0,0 +1,325 @@
...
+      ResultAggregator::CreateAggregator(
+          Benchmark::RepetitionModeE::MiddleHalfLoop);
+
+  Benchmark Result = std::move(AllResults[0]);
+
+  ResultAgg->AggregateResults(Result,
+                              ArrayRef<Benchmark>(AllResults).drop_front());
+
+  unsigned Throughput100 = static_cast<unsigned>(
+      round(Result.Measurements[0].PerSnippetValue * 100));
```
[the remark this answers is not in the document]
boomanaiden154: I've switched to returning the raw snippet value for now. We might want to make this a configurable in the future to enable easy comparisons with BHive datasets.

Also, do we need to do anything on the ML side for this change? I guess fp32 shouldn't have any problems adapting to the smaller values?

gematria/datasets/exegesis_benchmark.cc
boomanaiden154: I've added block and definition indices pretty much everywhere. Thanks for pushing for this.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -291,10 +309,13 @@ int main(int Argc, char *Argv[]) {
...
           BenchmarkRunner::ExecutionModeE::SubProcess, 30, {}, Benchmark::Min));
 
-  if (pfm::pfmInitialize())
-    ExitOnErr(llvm::make_error<StringError>(inconvertibleErrorCode(),
-                                            "Failed to initialize libpfm"));
+  auto JsonMemoryBuffer = ExitOnErr(
+      errorOrToExpected(MemoryBuffer::getFile(AnnotatedBlocksJson, true)));
+
+  auto ParsedAnnotatedBlocks =
+      ExitOnErr(json::parse(JsonMemoryBuffer->getBuffer()));
```
ondrasej, line 316: We should have a file error here too (IIUC, ``json::parse()` has no way of knowing the file name).

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -303,22 +324,35 @@ int main(int Argc, char *Argv[]) {
           errc::invalid_argument,
           "Malformed basic block: is not a JSON object"));
```
ondrasej, line 325: Same here. We should have the file name in the error message.

gematria/datasets/exegesis_benchmark.cc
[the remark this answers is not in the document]
ondrasej: On the ML side, we should be fine with any scaling, as long as the scaling is consistent. That said, I'd really go for normalizing to a single repetition, so that we don't need to worry about the number of repetitions of each data set (I already got burned by using different normalizations/scaling factors in the past, and I don't want to repeat that :))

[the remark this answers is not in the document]
boomanaiden154: Sounds good. The raw snippet value out of `llvm-exegesis` will already be normalized to a single repetition, so we'll be good there.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -291,10 +309,13 @@ int main(int Argc, char *Argv[]) {
...
-    ExitOnErr(llvm::make_error<StringError>(inconvertibleErrorCode(),
-                                            "Failed to initialize libpfm"));
+  auto JsonMemoryBuffer = ExitOnErr(
+      errorOrToExpected(MemoryBuffer::getFile(AnnotatedBlocksJson, true)));
+
+  auto ParsedAnnotatedBlocks =
+      ExitOnErr(json::parse(JsonMemoryBuffer->getBuffer()));
 
+  size_t BlockIndex = 0;
   for (const auto &AnnotatedBlock : *ParsedAnnotatedBlocks.getAsArray()) {
```
ondrasej: This would be easier with a regular for loop with an index since we have the index anyway. With the indexed for loop, we'd at least be safe from `continue`.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -303,22 +324,35 @@ int main(int Argc, char *Argv[]) {
...
                                  State, BenchCode.Key.LoopRegister);
 
-    unsigned Throughput100 =
-        ExitOnErr(benchmarkBasicBlock(BenchCode, *Runner, State));
+    Expected<double> ThroughputOrErr =
+        benchmarkBasicBlock(BenchCode, *Runner, State);
+
+    if (!ThroughputOrErr)
+      ExitOnErr(
+          createFileError(AnnotatedBlocksJson, ThroughputOrErr.takeError()));
```
ondrasej, line 352: Is there some LLVM function that would do this generically, i.e. would create the file error when there _was_ an error, and reutrn `success()` otherwise?

If there's none, it feels like there should be one :)

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -45,19 +45,23 @@ static cl::opt<std::string> AnnotatedBlocksJson(
...
+    const MCDisassembler &MachineDisassembler, const LLVMState &State,
+    size_t BlockIndex) {
   BenchmarkCode BenchCode;
 
   std::optional<StringRef> HexValue = BasicBlockJSON.getString("Hex");
   if (!HexValue)
     return llvm::make_error<StringError>(
-        errc::invalid_argument, "Expected basic block to have hex value");
+        errc::invalid_argument, "Malformed basic block: Basic block at index " +
+                                    Twine(BlockIndex) + " has no hex value");
```
ondrasej, line 56: The "Malformed basic block" part is repeated a couple of times. Given how lazy twines are, we can either pre-compute it - or move it into a lambda that can be called from here, so it doesn't need to be repeated.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -90,9 +96,11 @@ Expected<BenchmarkCode> parseJSONBlock(
...
   if (!RegisterDefinitions)
     return llvm::make_error<StringError>(
-        errc::invalid_argument,
-        "Malformed basic block: no register definitions array");
+        errc::invalid_argument, "Malformed basic block: Basic block at index " +
+                                    Twine(BlockIndex) +
+                                    " has no register definitions array");
 
+  size_t RegisterLoopIndex = 0;
   for (const auto &RegisterDefinitionValue : *RegisterDefinitions) {
```
ondrasej: Same comment as below for the blocks: wouldn't an index-based for loop be simpler here?

ondrasej (changes requested):

gematria/datasets/exegesis_benchmark.cc
boomanaiden154: There's one in `llvm-exegesis`. It's pretty generic. I don't think there would be too much opposition to moving it to `Error.h`. That would requite a bit of work though. For now I've added `ExitOnFileErr` locally.

gematria/datasets/exegesis_benchmark.cc
boomanaiden154: I added a `BasicBlockAtIndex` variable that holds that section and used it in all the errors.

gematria/datasets/exegesis_benchmark.cc
```diff
@@ -259,11 +255,30 @@ Expected<double> benchmarkBasicBlock(const BenchmarkCode &BenchCode,
...
+static void ExitOnFileError(const Twine &FileName, Error Err) {
+  if (Err) {
+    ExitOnErr(createFileError(FileName, std::move(Err)));
+  }
+}
+
+// Check E. If it's in a success state then return the contained value.
+// If it's in a failure state log the file error(s) and exit.
+template <typename T>
+T ExitOnFileError(const Twine &FileName, Expected<T> &&E) {
```
ondrasej, line 273: I guess these should be called `exitOnFileError()` to conform to the LLVM style function naming used in the rest of the file.

ondrasej (approved):
One more naming comment, otherwise it looks good. Feel free to merge once the comment is resolved.

PR #66, opened 2024-03-08
Add support for emitting loop register annotations

boomanaiden154:
This patch adds support for emitting loop register annotations for performing benchmarking in loop mode. Tests are added in addition to new flags to control the functionality where appropriate.


boomanaiden154:
Note that this is a stacked pull request on top of #51, so please don't merge it (if we get to that point) until #51 has already landed.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -261,6 +267,27 @@ int main(int argc, char* argv[]) {
...
+    }
+
+    // If we can't find a loop register, skip writing out this basic block
+    // so that downstream tooling doesn't execute the incorrect number of
+    // iterations.
+    if (!loop_register && skip_no_loop_register) {
+      std::cerr
+          << "Skipping block due to not being able to find a loop register\n";
+      continue;
+    }
```
ondrasej, line 289: I'd put this into a library so that it can be unit tested. IMO unit tests are a much better fit for this piece of code, because it is complex enough, and we're testing directly the loop counter detection behavior, not any accidental other behaviors.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -261,6 +267,27 @@ int main(int argc, char* argv[]) {
...
+      }
+    }
+
+    // If we can't find a loop register, skip writing out this basic block
+    // so that downstream tooling doesn't execute the incorrect number of
+    // iterations.
+    if (!loop_register && skip_no_loop_register) {
+      std::cerr
+          << "Skipping block due to not being able to find a loop register\n";
+      continue;
```
ondrasej, line 288: How often does this happen on the ComPile data set?

ondrasej (changes requested):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Refactored out into a library (`basic_block_utils` from a dependency PR) and unit-tested.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: I don't have any statistics on that currently. Will hopefully collect them once this script is working at scale.

gematria/datasets/basic_block_utils_test.cc
```diff
@@ -112,5 +118,38 @@ TEST_F(BasicBlockUtilsTest, UsedRegistersImplicitUse) {
   EXPECT_THAT(UsedRegisters, UnorderedElementsAre(X86::RAX, X86::RSP));
 }
 
+TEST_F(BasicBlockUtilsTest, LoopRegisterSingleInstruction) {
+  std::optional<unsigned> LoopRegister = getLoopRegister(R"asm(
+    mov %rax, %rcx
+  )asm");
+  EXPECT_EQ(*LoopRegister, X86::RDX);
```
ondrasej, line 125: This might break when LLVM changes register indices. That doesn't happen very often, but we've already been broken by this a couple of times.

A more reliable way would be to test that the returned value is one of the remaining GPRs.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
ondrasej: No problem. But we should count them and then report the total number at the end.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
ondrasej: 👍

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. The script now outputs information on this at the end.

gematria/datasets/basic_block_utils_test.cc
boomanaiden154: Interesting. Didn't realize LLVM would change the register indices that often. Good point though that this test would be brittle regardless. I've changed it as you've suggested.

gematria/datasets/basic_block_utils.h
```diff
@@ -28,19 +28,16 @@ namespace gematria {
...
-      const std::vector<DisassembledInstruction> &Instructions,
-      const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo);
-};
+// Gets the registers used by a basic block (in the form of a sequence of
+// instructions).
+std::vector<unsigned> getUsedRegisters(
+    const ArrayRef<DisassembledInstruction> Instructions,
+    const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo);
+
+std::optional<unsigned> getLoopRegister(
```
ondrasej, line 37: Please add a comment explaining what this does (returns a GPR that is not touched by the basic block). The name might sound like we're looking for some kind of counter used by the basic block.

It could also make sense to rename this to `getUnusedGPRegister()` to make it more clear what this does (instead of naming it by what it is used for).

ondrasej (changes requested):

ondrasej:
Please let me know once the parent PRs are merged and this is rebased, I'll review then.

boomanaiden154:
> Please let me know once the parent PRs are merged and this is rebased, I'll review then.

Everything has been rebased now.

gematria/datasets/basic_block_utils.cc
```diff
@@ -123,4 +123,53 @@ std::vector<unsigned> getUsedRegisters(
   return UsedRegistersList;
 }
 
+unsigned getSuperRegisterAllClasses(unsigned OriginalRegister,
+                                    const MCRegisterInfo &RegisterInfo) {
+  unsigned SuperRegister = OriginalRegister;
+  for (MCPhysReg CurrentSuperRegister :
+       RegisterInfo.superregs_inclusive(OriginalRegister))
+    SuperRegister = CurrentSuperRegister;
```
ondrasej, line 131: We should use the same check as in https://github.com/google/gematria/blob/ad44397422734b3e43765821b11842edc8c003e3/gematria/datasets/basic_block_utils.cc#L39 to make sure we're always moving towards the largest super register.

gematria/datasets/basic_block_utils.cc
```diff
@@ -123,4 +123,53 @@ std::vector<unsigned> getUsedRegisters(
...
+      }
+    }
+    for (unsigned ImplicitlyUsedRegister :
+         InstructionInfo.get(Instruction.mc_inst.getOpcode()).implicit_uses())
+      UsedRegisters.insert(
+          getSuperRegisterAllClasses(ImplicitlyUsedRegister, RegisterInfo));
+    for (unsigned ImplicitlyDefinedRegister :
+         InstructionInfo.get(Instruction.mc_inst.getOpcode()).implicit_defs())
+      UsedRegisters.insert(
+          getSuperRegisterAllClasses(ImplicitlyDefinedRegister, RegisterInfo));
```
ondrasej, line 157: Shorter:
```cc
const MCInstrDesc& Desc = InstructionInfo.get(Instruction.mc_inst.getOpcode());
UsedRegisters.insert(Desc.implicit_uses().begin(), Desc.implicit_uses.end());
UsedRegisters.insert(Desc.implicit_defs().begin(), Desc.implicit_defs.end());
```

ondrasej (changes requested):

PR #67, opened 2024-03-09
Add none annotator type

boomanaiden154:
This patch adds a none annotator type to
convert_bhive_to_llvm_exegesis_input that doesn't run any annotator. This is useful for looking at the output assembly for debugging purposes.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -49,10 +49,12 @@ constexpr std::string_view kMemDefPrefix = "# LLVM-EXEGESIS-MEM-DEF ";
 constexpr std::string_view kMemMapPrefix = "# LLVM-EXEGESIS-MEM-MAP ";
 constexpr std::string_view kMemNamePrefix = "MEM";
 
-enum class AnnotatorType { kExegesis, kFast };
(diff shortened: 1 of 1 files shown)


ondrasej:
While I have nothing against this change, why do we need this? If it is just to convert the hex strings into assembly, that can be done with llvm-mc:
```
echo "${bhive_hex}" | sed 's/\(..\)/0x\1, /g' | llvm-mc -disassemble -x86-asm-syntax=intel -mcpu=skx
```

ondrasej (approved):

boomanaiden154:
> While I have nothing against this change, why do we need this? If it is just to convert the hex strings into assembly, that can be done with llvm-mc:
> 
> ```
> echo "${bhive_hex}" | sed 's/\(..\)/0x\1, /g' | llvm-mc -disassemble -x86-asm-syntax=intel -mcpu=skx
> ```

Right, it can easily be done with `llvm-mc` (on top of other tools). Mainly, this makes it easy to test at scale on the exact format that is available. Finding issues where basic blocks won't disassemble with an extremely low frequency is going to be a pain to do with `llvm-mc`, and the annotators can further complicated things (like throwing unrelated errors due to other conditions). This also allows for inspection of specific BBs quite easily as one can just open up the relevant file after doing the disassembly.

PR #68, opened 2024-03-09
Add script to process and filter BBs

boomanaiden154:
Raw basic blocks extracted using information from the basic block address map need some processing before they are ready for benchmarking, on top of some filtering. For example, basic block termination instructions like jumps need to be removed. Other things need to be filtered like basic blocks containing call instructions or return instructions, or these instructions need to at least be removed. This PR adds a script designed to do these things, initially aimed at removing terminator instructions.


gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,87 @@
...
+
+static cl::opt<std::string> OutputFile(
+    "output-file",
+    cl::desc(
+        "Path to the output CSV file with processed/filtered basic blocks"),
+    cl::init(""));
+
+static cl::opt<bool> LastInstructionIsTerminator(
+    "last-instruction-is-terminator",
+    cl::desc("Whether or not the last instruction changes the control flow"),
```
ondrasej, line 39: This should be written in the form of an action, e.g. "Remove the last instruction of each basic block, if it is an instruction that changes the control flow." or "Filter out blocks that have a control-flow-changing instruction as the last instruction in the block."

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,87 @@
...
+    cl::init(true));
+
+static ExitOnError ExitOnErr("process_and_filter_bbs error: ");
+
+static void ExitOnFileError(const Twine &FileName, Error Err) {
+  if (Err) {
+    ExitOnErr(createFileError(FileName, std::move(Err)));
+  }
+}
```
ondrasej, line 48: Same comment as in #65: I'd move this into `main()`, and also move most of the functionality out of `main()` into a function that returns an error or a file error.

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,87 @@
...
+
+  std::unique_ptr<MCInstPrinter> MachineInstructionPrinter =
+      llvm_support->CreateMCInstPrinter(0);
+
+  std::ifstream InputFileStream(InputFile);
+  std::ofstream OutputFileStream(OutputFile);
+  for (std::string Line; std::getline(InputFileStream, Line);) {
+    auto MachineCodeHex = gematria::ParseHexString(Line);
+    if (!MachineCodeHex.has_value()) {
+      dbgs() << "could not parse: " << Line << "\n";
```
ondrasej, line 64: Exit with a file error?

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,87 @@
...
+      return 3;
+    }
+
+    std::vector<gematria::DisassembledInstruction> DisassembledInstructions =
+        ExitOnErr(gematria::DisassembleAllInstructions(
```
ondrasej, line 69: Exit with a file error?

ondrasej (changes requested):

gematria/datasets/process_and_filter_bbs.cc
boomanaiden154: This is already a dead option (sorry for the confusion here). I originally had it setup to just remove the last instruction, but it's easier to just skip branches/calls/returns in general. I had already switched to that but forgot to remove the option.

gematria/datasets/process_and_filter_bbs.cc
boomanaiden154: Done. Moved out the functionality for processing individual blocks into a function called `ProcessBasicBlock` that cleans up the error handling. Thanks for the suggestion!

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,97 @@
...
+    cl::desc(
+        "Path to the output CSV file with processed/filtered basic blocks"),
+    cl::init(""));
+
+static ExitOnError ExitOnErr("process_and_filter_bbs error: ");
+
+Expected<std::string> ProcessBasicBlock(
+    const std::string &BasicBlock,
+    const std::unique_ptr<gematria::LlvmArchitectureSupport> &llvm_support,
+    const std::unique_ptr<MCInstPrinter> &MachineInstructionPrinter,
```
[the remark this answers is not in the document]
boomanaiden154: Good point. Switched.

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,96 @@
...
+    cl::desc("Path to the input CSV file containing hex basic blocks"),
+    cl::init(""));
+
+static cl::opt<std::string> OutputFile(
+    "output-file",
+    cl::desc(
+        "Path to the output CSV file with processed/filtered basic blocks"),
+    cl::init(""));
+
+Expected<std::string> ProcessBasicBlock(
```
ondrasej, line 37: Very nitpicky: `processBasicBlock` to be in line with the LLVM naming conventions.

ondrasej (approved):

PR #69, opened 2024-03-09
Fix clang-formatting issue

boomanaiden154:
This patch fixes a clang-formatting issue introduced in a previous patch.


ondrasej (commented):
This doesn't look like a clang-format problem to me (the single quotes are part of the output, they surround the basic block hex string).

boomanaiden154:
Going to merge this to at least get the clang-format issue fixed at tip of tree. Will investigate the Bazel CI failures further.

PR #70, opened 2024-03-09
Add support for filtering call and return instructions

boomanaiden154:
This patch adds support for filtering return and call instructions out of basic blocks, which make them impossible to benchmark due to a lack of surrounding context.


boomanaiden154:
Closing this, will integrate it into the original PR.

PR #71, opened 2024-03-12
Add basic block utils library with function to calculate register usage

boomanaiden154:
This patch adds a new `basic_block_utils` library that will be extended in subsequent patches to add additional functionality necessary for processing datasets of basic blocks. Currently, `BasicBlockUtils::getUsedRegisters` is implemented for #51.

### DIFF
--- a/gematria/datasets/basic_block_utils.cc
+++ b/gematria/datasets/basic_block_utils.cc
@@ -1,0 +1,54 @@
+// Copyright 2024 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <map>
+
+#include "gematria/datasets/basic_block_utils.h"
+#include "llvm/MC/MCRegisterInfo.h"
+#include "llvm/MC/MCInstrInfo.h"
+
+using namespace llvm;
+
+namespace gematria {
+
+std::vector<unsigned> BasicBlockUtils::getUsedRegisters(const std::vector<DisassembledInstruction> &Instructions, const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo) {
+  std::map<unsigned, bool> UsedRegisters;
+  for (const gematria::DisassembledInstruction& Instruction : Instructions) {
+    for (unsigned OperandIndex = 0;
+         OperandIndex < Instruction.mc_inst.getNumOperands();
+         ++OperandIndex) {
+      if (Instruction.mc_inst.getOperand(OperandIndex).isReg()) {
+        unsigned RegisterNumber =
+            Instruction.mc_inst.getOperand(OperandIndex).getReg();
+        if (RegisterNumber == 0) continue;
+        unsigned SuperRegisterNumber = 0;
+        for (MCPhysReg CurrentSuperRegister :
+             RegisterInfo.superregs_inclusive(RegisterNumber)) {
+          SuperRegisterNumber = CurrentSuperRegister;
+        }
+        UsedRegisters[SuperRegisterNumber] = true;
+      }
+    }
+  }
+
+  std::vector<unsigned> UsedRegistersList;
+  UsedRegistersList.reserve(UsedRegisters.size());
+
+  for (const auto [RegisterIndex, RegisterUsed] : UsedRegisters)
+    UsedRegistersList.push_back(RegisterIndex);
+
+  return UsedRegistersList;
+}
+
+} // namespace gematria
--- a/gematria/datasets/basic_block_utils.cc
+++ b/gematria/datasets/basic_block_utils.cc
@@ -12,22 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "gematria/datasets/basic_block_utils.h"
+
 #include <map>
 
-#include "gematria/datasets/basic_block_utils.h"
-#include "llvm/MC/MCRegisterInfo.h"
 #include "llvm/MC/MCInstrInfo.h"
+#include "llvm/MC/MCRegisterInfo.h"
 
 using namespace llvm;
 
 namespace gematria {
 
-std::vector<unsigned> BasicBlockUtils::getUsedRegisters(const std::vector<DisassembledInstruction> &Instructions, const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo) {
+std::vector<unsigned> BasicBlockUtils::getUsedRegisters(
+    const std::vector<DisassembledInstruction> &Instructions,
+    const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo) {
   std::map<unsigned, bool> UsedRegisters;
-  for (const gematria::DisassembledInstruction& Instruction : Instructions) {
+  for (const gematria::DisassembledInstruction &Instruction : Instructions) {
     for (unsigned OperandIndex = 0;
-         OperandIndex < Instruction.mc_inst.getNumOperands();
-         ++OperandIndex) {
+         OperandIndex < Instruction.mc_inst.getNumOperands(); ++OperandIndex) {
       if (Instruction.mc_inst.getOperand(OperandIndex).isReg()) {
         unsigned RegisterNumber =
             Instruction.mc_inst.getOperand(OperandIndex).getReg();
@@ -51,4 +53,4 @@ std::vector<unsigned> BasicBlockUtils::getUsedRegisters(const std::vector<Disass
   return UsedRegistersList;
 }
 
-} // namespace gematria
+}  // namespace gematria
--- a/gematria/datasets/basic_block_utils_test.cc
+++ b/gematria/datasets/basic_block_utils_test.cc
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "X86.h"
+#include "X86InstrInfo.h"
+#include "X86RegisterInfo.h"
 #include "gematria/datasets/basic_block_utils.h"
 #include "gematria/llvm/llvm_architecture_support.h"
 #include "gematria/llvm/asm_parser.h"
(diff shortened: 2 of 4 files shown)


boomanaiden154:
This is part of a PR stack, being a dependency of #51.

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+unsigned getSuperRegister(unsigned OriginalRegister,
+                          const MCRegisterInfo &RegisterInfo) {
+  unsigned SuperRegister = OriginalRegister;
+  for (MCPhysReg CurrentSuperRegister :
+       RegisterInfo.superregs_inclusive(OriginalRegister)) {
+    SuperRegister = CurrentSuperRegister;
+  }
+  // Only return super registers for GPRs. Since mainly GPRs will be used for
+  // addressing, redefining other aliasing registers (like vector registers)
+  // does not matter as much.
```
ondrasej, line 38: For others (in particular for XMMs), this logic is fine, but we might want to post-process it afterwards, and if there are overlapping vector registers, take just the largest one.

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+    return SuperRegister;
+  else
+    return OriginalRegister;
+}
+
+std::vector<unsigned> BasicBlockUtils::getUsedRegisters(
+    const std::vector<DisassembledInstruction> &Instructions,
+    const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo) {
+  std::map<unsigned, bool> UsedInputRegisters;
+  std::map<unsigned, bool> DefinedInBBRegisters;
```
ondrasej, line 49: These two are defined as maps, but they are used as sets. It would be more efficient to use `std::set`.

But given that the number of registers is fairly small, it might be even faster to use `llvm::BitVector` that has a very nice cache profile and very fast access (ideally, `std::vector<bool>` but that doesn't have a good API for iterating over all ones).

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+            true;
+      }
+    }
+  }
+
+  std::vector<unsigned> UsedRegistersList;
+  UsedRegistersList.reserve(UsedInputRegisters.size());
+
+  for (const auto [RegisterIndex, RegisterUsed] : UsedInputRegisters)
+    UsedRegistersList.push_back(RegisterIndex);
```
ondrasej, line 88: Shorter:
```
UsedRegistersList.assign(UsedInputRegisters.begin(), UsedInputRegistes.end());
```
or, with `llvm::BitVector`,
```
UsedRegistersList.assign(UsedInputRegisters.set_bits_begin(), UsedInputRegistes.set_bits_end());
```

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+    const std::vector<DisassembledInstruction> &Instructions,
+    const MCRegisterInfo &RegisterInfo, const MCInstrInfo &InstructionInfo) {
+  std::map<unsigned, bool> UsedInputRegisters;
+  std::map<unsigned, bool> DefinedInBBRegisters;
+  for (const gematria::DisassembledInstruction &Instruction : Instructions) {
+    unsigned InstructionDefs =
+        InstructionInfo.get(Instruction.mc_inst.getOpcode()).getNumDefs();
+    for (unsigned OperandIndex = InstructionDefs;
+         OperandIndex < Instruction.mc_inst.getNumOperands(); ++OperandIndex) {
+      if (Instruction.mc_inst.getOperand(OperandIndex).isReg()) {
```
ondrasej, line 55: Very nit: Above
```
const auto& Operand = Instruction.mc_inst.getOperand(OperandIndex);
```
and then just use `Operand`.

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+        if (DefinedInBBRegisters.count(SuperRegister) > 0) continue;
+        UsedInputRegisters[SuperRegister] = true;
+      }
+    }
+    // We also need to handle instructions that have implict uses.
+    for (unsigned ImplicitlyUsedRegister :
+         InstructionInfo.get(Instruction.mc_inst.getOpcode()).implicit_uses()) {
+      UsedInputRegisters[getSuperRegister(ImplicitlyUsedRegister,
+                                          RegisterInfo)] = true;
+    }
```
ondrasej, line 71: We also need to track implicit defs of the register.

gematria/datasets/basic_block_utils_test.cc
```diff
@@ -0,0 +1,116 @@
...
+    return Instructions;
+  }
+
+  std::vector<unsigned> getUsedRegisters(std::string_view TextualAssembly) {
+    return BasicBlockUtils::getUsedRegisters(
+        getInstructions(TextualAssembly), LlvmArchSupport->mc_register_info(),
+        LlvmArchSupport->mc_instr_info());
+  }
+};
+
```
ondrasej, line 61: A couple of other fun cases to test, maybe with some instructions before and after to make sure that the uses and defs are tracked correctly.

1. Lots of implicit uses and defs:
```asm
cpuid
```
2. `%ebx` is both a use and a def, implicit def of `%eflags`:
```asm
add %eax, %ebx
```
3. register aliasing is so much fun:
```asm
mov %al, 1      ; Write to a small part of RAX, rest is preserved.
add %eax, %ebx  ; As weird as it looks, RAX is still an input to the block.
```
4. register aliasing is even more fun:
```asm
mov %eax, 1     ; Cleans the upper bits of RAX (consistency FTW)
add %rax, %rbx  ; RAX is not an input of the block.
```
4. Addressing modes use registers too, let's make sure they are accounted for
```asm
add %rax, -16(%rbx, %rcx, 8)
```
6. Implicit use of `%eflags`, explicit output
```asm
seta %al
```
7. Implicit uses, implicit defs:
```asm
rep movsb
```

ondrasej (changes requested):
Thanks for taking time to clean this up!

gematria/datasets/basic_block_utils.cc
boomanaiden154: That's a good point. I'll look into doing that at some point. My main concern with immediately bumping all the vector registers too was that it is super easy to end up registers that aren't supported by the current CPU without looking at what is supported by the target (which is reasonably complicated logic). This works for now and should be adjustable in the future.

gematria/datasets/basic_block_utils.cc
```diff
@@ -0,0 +1,93 @@
...
+
+namespace gematria {
+
+unsigned getSuperRegister(unsigned OriginalRegister,
+                          const MCRegisterInfo &RegisterInfo) {
+  unsigned SuperRegister = OriginalRegister;
+  for (MCPhysReg CurrentSuperRegister :
+       RegisterInfo.superregs_inclusive(OriginalRegister)) {
+    SuperRegister = CurrentSuperRegister;
+  }
```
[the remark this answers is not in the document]
boomanaiden154: It will always return a super-register (at least from what I understand). So if we give it `AL`, it won't return `AX`, it would just return `EAX`. Just pulling the back of the iterator returns zero, which seems like an issue on the LLVM side. I'll add a TODO.

gematria/datasets/basic_block_utils.cc
boomanaiden154: Switch to `std::set` for now.

gematria/datasets/basic_block_utils.cc
boomanaiden154: Done. Test added with `rdtsc`.

gematria/datasets/basic_block_utils_test.cc
boomanaiden154: Added most of these in (and fixed issues with register aliasing). I didn't add the last two as number six is just testing use of the flag register, which we're not accounting for, and `rep movsb` is kind of redundant with testing for `cpuid`/`rdtsc`.

gematria/datasets/basic_block_utils.cc
```diff
@@ -42,50 +47,75 @@ unsigned getSuperRegister(unsigned OriginalRegister,
...
+      // already being defined, as moving values into those registers does not
+      // zero the rest of the bits in the register, unlike when writing to
+      // 32-bit registers.
+      if (RegisterInfo.getRegClass(X86::GR8RegClassID)
+              .contains(ImplicitlyDefinedRegister) ||
+          RegisterInfo.getRegClass(X86::GR16RegClassID)
+              .contains(ImplicitlyDefinedRegister))
+        continue;
+      DefinedInBBRegisters.insert(
+          getSuperRegister(ImplicitlyDefinedRegister, RegisterInfo));
```
ondrasej, line 91: Nitpicky: this is repeated almost verbatim below, let's put it e.g. to a lambda and then just call the lambda.

ondrasej (changes requested):

gematria/datasets/basic_block_utils.cc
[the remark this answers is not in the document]
boomanaiden154: It will return `RAX`. I misspoke there.

Looking over the `superregs_inclusive` iterator implementation again, it doesn't look like each successive register is guaranteed to be a super register. I've added a check that the register we're assigning to `SuperRegister` is in fact a super register of it.

I would add an assertion that the bit width is bigger, but it seems surprisingly hard to get the bit width. One either needs to go through `TargetRegisterInfo` (which seems to be a pain to get here), or get the register class, but the register class is not guaranteed to have the bit width (although I would think it would for X86 GPRs), and there is seemingly no direct mapping from registers to their register classes in `MCRegisterInfo`.

Either way, the explicit super register check should be fine.

gematria/datasets/basic_block_utils_test.cc
```diff
@@ -92,25 +93,76 @@ TEST_F(BasicBlockUtilsTest, UsedRegistersMultipleRegisters) {
...
+  std::vector<unsigned> UsedRegisters = getUsedRegs(R"asm(
+    addq %rax, %rbx
+  )asm");
+  EXPECT_THAT(UsedRegisters, UnorderedElementsAre(X86::RAX, X86::RBX));
+}
+
+TEST_F(BasicBlockUtilsTest, UsedRegistersRegisterAliasing) {
+  std::vector<unsigned> UsedRegisters = getUsedRegs(R"asm(
+    movb 1, %al
+    addq %rax, %rbx
```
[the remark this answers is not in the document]
boomanaiden154: Looks like it returns `%RCX` as well. I'll look into the issue at some point. For now I've disabled the test and added a todo.

ondrasej (approved):

gematria/datasets/basic_block_utils.cc
```diff
@@ -36,17 +36,35 @@ unsigned getSuperRegister(unsigned OriginalRegister,
...
+      SuperRegister = CurrentSuperRegister;
+    }
   }
   // Only return super registers for GPRs. Since mainly GPRs will be used for
   // addressing, redefining other aliasing registers (like vector registers)
   // does not matter as much.
+  // TODO(boomanaiden154): We are only handling the simple case as it gives the
+  // most mileage, and vector registers need additional target-specific handling
+  // to ensure that the instructions are actually supported by the CPU we are
+  // executing on. This should be fixed in the future.
```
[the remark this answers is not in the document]
boomanaiden154: That would make quite a bit of sense. I will look into that at some point in the future.

gematria/datasets/BUILD.bazel
```diff
@@ -228,4 +228,24 @@ cc_test(
         "//gematria/testing:matchers",
         "@com_google_googletest//:gtest_main",
     ],
+cc_binary(
```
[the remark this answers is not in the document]
boomanaiden154: Yep, a bad merge that I somehow forgot to cleanup. After the cleanup, the diff isn't there.

PR #73, opened 2024-03-14
Make exegesis annotator error out when remapping an address

boomanaiden154:
Currently, there is a failure condition seen in some blocks where the annotator finds a segfault, tells exegesis to map a page there, and then the block fails again at the same address. This needs some debugging, but currently it will be helpful if we can just log a warning message and skip this basic block.

### DIFF
--- a/gematria/datasets/find_accessed_addrs_exegesis.cc
+++ b/gematria/datasets/find_accessed_addrs_exegesis.cc
@@ -152,6 +152,9 @@ Expected<AccessedAddrs> ExegesisAnnotator::findAccessedAddrs(
           uintptr_t MapAddress = (CrashInfo.getAddress() & ~0xfff);
           if (MapAddress == 0)
             return make_error<Failure>("Segfault at zero address, cannot map.");
+          // TODO(boomanaiden154): The fault captured below occurs when
+	  // exegesis tries to map an address and the mmap fails. When these
+	  // errors are handled within exegesis, we should remove this check.
           for (const auto &MemoryMapping : BenchCode.Key.MemoryMappings) {
             if (MemoryMapping.Address == MapAddress)
               return make_error<Failure>(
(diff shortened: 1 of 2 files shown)


boomanaiden154:
This is a stacked PR, depending on #72, mostly so that I can avoid merge conflicts when cherry-picking commits from a running branch I have with a couple fixes.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -152,6 +152,11 @@ Expected<AccessedAddrs> ExegesisAnnotator::findAccessedAddrs(
           uintptr_t MapAddress = (CrashInfo.getAddress() & ~0xfff);
           if (MapAddress == 0)
             return make_error<Failure>("Segfault at zero address, cannot map.");
+          for (const auto &MemoryMapping : BenchCode.Key.MemoryMappings) {
+            if (MemoryMapping.Address == MapAddress)
+              return make_error<Failure>(
+                  "Segfault at an address already mapped.");
+          }
```
ondrasej, line 159: What kind of error happens here? Do I understand correctly that this happens when `mmap()` fails to map the memory we asked it to map?

If yes, can we catch this earlier? E.g. during the mmaping phase?

boomanaiden154: I haven't done too much investigation (yet), but that's what seems to be happening. I agree that exegesis needs correct error handling on `mmap` failures, I just haven't had time to implement since it's relatively complicated (requires writing code to manually generate the assembly for the error handling). It's on my TODO list, but somewhat low priority (will get bumped up with this issue).

ondrasej: SG, then please add a TODO note also here.

ondrasej (approved):

PR #74, opened 2024-03-14
Make the annotator return an error if there are too many annotations

boomanaiden154:
This patch adds a threshold for the number of annotations that the exegesis annotator will make before erroring out. Otherwise, blocks that for example, access a new page on every iteration, will need 5000 different annotations which will take quite a while to annotate.


boomanaiden154:
This is also a stacked PR with a dependency of #73, again to avoid merge conflicts when cherry-picking.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -32,6 +32,10 @@ using namespace llvm::exegesis;
 
 namespace gematria {
 
+// This value specifies the maximum number of times the annotator will attempt
+// to map segmentation fault addresses before giving up and throwing an error.
+constexpr unsigned kMaxAnnotationAttempts = 100;
+
```
ondrasej, line 38: If it doesn't complicate things too much, this should be a parameter that can be tuned (from the command-line).

ondrasej (changes requested):

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Done. It's now a parameter in the BHive to Exegesis conversion script.

ondrasej (approved):

PR #77, opened 2024-03-17
Add functionality to convert JSON datasets to Gematria protos.

virajbshah:
* JSON datasets, while not very useful otherwise, being easy to read are handy for debugging while also being rich enough to store extra, possibly nested, data such as instruction annotations.
 * Note: the current JSON format includes instruction annotations and so depends on `AnnotationProto` from #27, and will not build without `gematria/proto/annotation.proto`.

### DIFF
--- a/gematria/datasets/BUILD
+++ b/gematria/datasets/BUILD
@@ -32,6 +32,24 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "json_importer",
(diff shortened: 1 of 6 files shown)


virajbshah:
It makes more sense to use the Gematria `basic_block` data structures and the proto format over something like JSON, so this is being closed.

PR #80, opened 2024-03-24
Refactor asm output into separate function

boomanaiden154:
This patch refactors the asm output in the annotation script into a separate function. This is intended to make further work on the code, like implementing threading, easier in the future.

PR #81, opened 2024-03-24
Refactor JSON snippet construction into sepaarate function

boomanaiden154:
This patch refactors the construction of a JSON representation of an annotated snippet into a separate function. This is intended to make future work, like implementing threading, easier.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -186,6 +186,57 @@ absl::StatusOr<AnnotatedBlock> AnnotateBasicBlock(
   return std::move(annotated_block);
 }
 
+llvm::json::Value GetJSONForSnippet(const AnnotatedBlock& annotated_block,
+                                    std::string_view hex) {
+  llvm::json::Object current_snippet;
+
+  llvm::json::Array register_definitions;
+  for (const auto register_id : annotated_block.used_registers) {
+    llvm::json::Object current_register_definition;
+    current_register_definition["Register"] = register_id;
+    current_register_definition["Value"] = kInitialRegVal;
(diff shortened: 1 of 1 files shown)


boomanaiden154:
This is stacked on https://github.com/google/gematria/pull/80.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -186,6 +186,57 @@ absl::StatusOr<AnnotatedBlock> AnnotateBasicBlock(
...
+    register_definitions.push_back(std::move(current_register_definition));
+  }
+  current_snippet["RegisterDefinitions"] =
+      llvm::json::Value(std::move(register_definitions));
+
+  // Output the loop register.
+  if (annotated_block.loop_register)
+    current_snippet["LoopRegister"] = *annotated_block.loop_register;
+  else
+    current_snippet["LoopRegister"] = llvm::json::Value(nullptr);
```
ondrasej, line 207: What does this do in terms of the JSON output, insert `null`?

If yes, we might be better off with skipping the key (or using e.g. `0` or `-1` as a sentinel value), because some JSON tools and libraries do not seem to consider `null` to be a valid JSON value, and choke on it.

ondrasej (changes requested):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Switched to using `llvm::MCRegister::NoRegister`.

ondrasej (commented):
LGTM. I'll approve once #80 is merged (and this is rebased on top of it).

ondrasej (approved):

PR #82, opened 2024-03-24
Change default annotator implementation to exegesis

boomanaiden154:
This patch changes the default annotator implementation to exegesis. This is due to the fast annotator deadlocking when used with future patches that utilize threding to speed up the annotation process on large datasets. This casues tests to randomly deadlock with the fast annotator. Instead of manually changing each individual test, just adjust the default.


boomanaiden154:
This is stacked on https://github.com/google/gematria/pull/81.

boomanaiden154:
Closing as this was only really necessary for #83.

PR #83, opened 2024-03-24
Parallelize memory annotations

### LINKED ISSUE #56: Parallelize memory annotations
The current script in `./gematria/datasets/convert_bhive_to_exegesis_inputs.cc` runs sequentially. This is somewhat of a problem for using the Exegesis annotator, which isn't particularly fast. This can easily be parallelized as we don't care about the timings at all while running the annotations. This should be doable with some refactoring and use of LLVM's threading APIs.

boomanaiden154:
This patch parallelizes memory annotations in the
convert_bhive_to_llvm_exegesis script, particularly to speed up annotations using the exegesis annotator. This uses the LLVM threading primitives, particularly ThreadPool.

This fixes #56.


boomanaiden154:
This is stacked on https://github.com/google/gematria/pull/82.

boomanaiden154:
Closing this as the threading doesn't actually end up improving performance. I've spent a bit of time debugging, but was never able to figure it out. Using an external script for orchestration with multiple processes (python+ray) sped things up about linearly with the number of cores.

PR #87, opened 2024-03-25
Add tests for extract_bbs_from_obj

boomanaiden154:
This patch adds testing infrastructure and initial tests for the extract_bbs_from_obj script to ensure that all functionality is working.


boomanaiden154:
This is stacked on https://github.com/google/gematria/pull/86.

gematria/datasets/extract_bbs_from_obj_test.py
```diff
@@ -0,0 +1,11 @@
+from lit.main import main
```
ondrasej, line 1: Nit: The Google Python style does not allow direct imports of classes and functions.

gematria/datasets/extract_bbs_from_obj_test.py
ondrasej, line 1: Please add the Apache 2.0 license notice. Same for all the other Python files.

ondrasej (changes requested):

gematria/datasets/extract_bbs_from_obj_test.py
boomanaiden154: Added.

gematria/datasets/extract_bbs_from_obj_test.py
boomanaiden154: Fixed.

ondrasej (approved):

PR #90, opened 2024-03-26
Add support for filtering blocks that access memory

boomanaiden154:
This patch adds an option that enables filtering blocks that access memory. This is intended for testing as memory-accessing blocks can be harder to benchmark in terms of noise, so comparing with and without memory accessing blocks might be useful.


boomanaiden154:
Note that this is stacked on #68.

ondrasej (approved):

PR #92, opened 2024-03-26
Add instruction annotations to the basic block representations.

virajbshah:
Adds instruction annotations to the proto specification (`AnnotationProto`) , the graph builder basic block representation (`Annotation`), and functions to convert between the two (`AnnotationFromProto` and `ProtoFromAnnotation`).

### DIFF
--- a/gematria/basic_block/basic_block_protos.cc
+++ b/gematria/basic_block/basic_block_protos.cc
@@ -20,10 +20,36 @@
 #include <vector>
 
 #include "gematria/basic_block/basic_block.h"
+#include "gematria/proto/annotation.pb.h"
 #include "gematria/proto/canonicalized_instruction.pb.h"
 #include "google/protobuf/repeated_ptr_field.h"
 
 namespace gematria {
+
+namespace {
+
+template <typename Object, typename Proto, typename Convertor>
+std::vector<Object> ToVector(
+    const google::protobuf::RepeatedPtrField<Proto>& protos,
+    Convertor object_from_proto) {
+  std::vector<Object> result(protos.size());
+  std::transform(protos.begin(), protos.end(), result.begin(),
+                 object_from_proto);
+  return result;
+}
+
+template <typename Object, typename Proto, typename Convertor>
+void ToRepeatedPtrField(
+    const std::vector<Object>& objects,
+    google::protobuf::RepeatedPtrField<Proto>* repeated_field,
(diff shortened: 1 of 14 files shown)


gematria/basic_block/basic_block_protos.cc
```diff
@@ -20,11 +20,37 @@
...
 namespace gematria {
 
+namespace {
+
+template <typename Object, typename Proto, typename Convertor>
+std::vector<Object> ToVector(
+    const google::protobuf::RepeatedPtrField<Proto>& protos,
+    Convertor object_from_proto) {
+  std::vector<Object> result(protos.size());
+  std::transform(protos.begin(), protos.end(), result.begin(),
```
ondrasej, line 36: This is very nitpicky, but since we're here making this generic, let's go with `std::begin()`, `std::end()`, and `std::size()` instead of calling the methods directly.

gematria/basic_block/basic_block_protos.cc
```diff
@@ -20,11 +20,37 @@
...
+  return result;
+}
+
+template <typename Object, typename Proto, typename Convertor>
+void ToRepeatedPtrField(
+    const std::vector<Object>& objects,
+    google::protobuf::RepeatedPtrField<Proto>* repeated_field,
+    Convertor proto_from_object) {
+  repeated_field->Reserve(objects.size());
+  std::transform(objects.begin(), objects.end(),
```
ondrasej, line 47: Same here.

ondrasej (approved):
This looks good. I've added a few minor comments, but feel free to merge once they are resolved (or let me know and I'll merge).

gematria/basic_block/python/basic_block.cc
```diff
@@ -185,24 +186,44 @@ PYBIND11_MODULE(basic_block, m) {
 
   py::bind_vector<std::vector<InstructionOperand>>(m, "InstructionOperandList");
 
-  py::class_<Instruction>(m, "Instruction")
+  py::class_<Annotation> annotation(m, "Annotation");
+  annotation
+      .def(py::init<std::string /* name */, double /* value */>(),
+           py::arg("name") = std::string(), py::arg("value"))
```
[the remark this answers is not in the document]
virajbshah: Right, that must've slipped past while removing name-only annotations and fiddling the default annotation value.

ondrasej (approved):

PR #93, opened 2024-03-26
Add support for instruction annotations to the graph builder.

virajbshah:
Allows `BasicBlockGraphBuilder` to store instruction annotations in a vector-of-vector matrix while processing instructions to build graphs, so that they are available to graph builder based models. The list of names of annotation types to be stored is supplied at the time of construction of the graph builder.


gematria/granite/python/graph_builder_test.py
ondrasej, line 1: Please update also `BasicBlockBuilderTest.assertBuilderIsSelfConsistent()` (that mainly checks dimensions of all the generated vectors and matrices, so we should have the same for the annotations).

gematria/testing/testdata/annotated_basic_blocks_with_throughput.pbtxt
ondrasej, line 1: Where do the blocks come from? Do we need to worry about licenses/origins/...?

ondrasej (changes requested):

gematria/testing/testdata/annotated_basic_blocks_with_throughput.pbtxt
virajbshah: They are the same blocks as those from `basic_blocks_with_throughput.pbtxt`, just with (random) annotations - so I don't think there should be any concerns.

PR #94, opened 2024-03-26
Add support for consuming instruction annotations to `TokenGraphBuilderModel`s. 

virajbshah:
Uses instruction annotations obtained from the graph builder to compute instruction node embeddings. The ends of the instruction node embedding vectors are overwritten by the values of the annotations on the corresponding instruction.


gematria/granite/python/graph_builder_model_base.py
```diff
@@ -187,11 +190,13 @@ def _create_tf_graph(self) -> None:
...
-        shape=(None,),
-        name=GraphBuilderModelBase.INSTRUCTION_NODE_MASK_TENSOR_NAME,
-    )
+    # May already be created by subclass.
+    if self._instruction_node_mask is None:
+      self._instruction_node_mask = tf.placeholder(
+          dtype=tf.dtypes.bool,
+          shape=(None,),
+          name=GraphBuilderModelBase.INSTRUCTION_NODE_MASK_TENSOR_NAME,
+      )
```
ondrasej, line 199: Under what circumstances would it be created by a subclass? (and does the subclass need to do that?)

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -50,6 +52,20 @@ class TokenGraphBuilderModel(graph_builder_model_base.GraphBuilderModelBase):
...
+
+  # A 1D byte tensor that contains the list of annotation names in the order of
+  # their indices in the graph builder.
+  _annotation_name_tensor: tf.Tensor
+
+  # The list of annotation names, in the order of their indices in the model.
+  _annotation_name_list: Sequence[str]
+
+  _instruction_annotations: tf.Tensor
+
```
ondrasej, line 68: Since this is already in the base graph builder model, we should probably move all these definitions to that class (and make the annotations available in any graph builder model).

Eventually (but definitely not in this PR chain), we can probably merge `GraphBuilderModelBase` and `TokenGraphBuilderModel` because there is no other graph builder model anyway.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -340,3 +398,72 @@ def _create_graph_network_modules(
...
+    if self.num_annotations == 0:
+      return embeddings
+
+    out = tf.concat(
+        [
+            tf.slice(
+                embeddings,
+                begin=[0, 0],
+                size=[-1, self.embed_dim - self.num_annotations],
```
ondrasej, line 455: Ideally, we should put this to the model (i.e. either compute the learned embedding size from a specified node feature size + annotation size, or compute the node state size from a specified learned embedding size + annotation size).

ondrasej (changes requested):

gematria/granite/python/graph_builder_model_base.py
virajbshah: It would always be created by `TokenGraphBuilderModel`s - to figure out what nodes are instructions and to update their embeddings with annotations. We don't really need this since we won't be making instances of `GraphBuilderModelBase`s, but I left it in to ensure it is there in case there is ever another subclass that doesn't create it.

Another option is to create it within `GraphBuilderModelBase` earlier, so that it is available to `TokenGraphBuilderModel`s when required.

gematria/model/python/token_model_flags.py
```diff
@@ -93,6 +105,38 @@ def get_oov_token_behavior_from_command_line_flags() -> (
   )
 
 
+def get_lines_from_command_line_flag(
+    flag: flags.FlagHolder[Text | None],
```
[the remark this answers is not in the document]
virajbshah: No reason particularly - passing in the file name sounds like a better idea. I'll do that.

gematria/granite/python/graph_builder_model_base.py
virajbshah: This will be removed since moving the creation of `_instruction_node_mask` should be done alongside the creation of annotation-related tensors which is moving to `GraphBuilderModelBase` as suggested in another comment.

gematria/granite/python/token_graph_builder_model.py
virajbshah: Sorry, I'm not sure I understand what is to be changed - this snippet should update the last `num_annotations` with the annotations for instruction nodes and keep the rest `embed_dim - num_annotations` intact, with `embed_dim` and `num_annotations` both coming from the model.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -340,3 +368,72 @@ def _create_graph_network_modules(
...
+            ),
+            tf.tensor_scatter_nd_update(
+                tf.slice(
+                    embeddings,
+                    begin=[0, self.embed_dim - self.num_annotations],
+                    size=[-1, self.num_annotations],
+                ),
+                indices=tf.where(self.instruction_node_mask),
+                updates=self.instruction_annotations,
+            ),
```
ondrasej, line 435: Did you test this with TF inference? The set of supported ops in TFLite is a bit limited, and more specialized ones (like scatters and gathers) might not be there.

If not, please add a TODO here and we should test it when possible (and maybe look for another way to implement it).

gematria/model/python/token_model_flags.py
```diff
@@ -38,6 +37,18 @@
     ),
 )
 
+_ANNOTATION_NAME_FILE = flags.DEFINE_string(
```
ondrasej, line 40: Very nitpicky: `_ANNOTATION_NAMES_FILE` to match the name of the command-line flag.

gematria/model/python/token_model_flags.py
```diff
@@ -114,25 +157,40 @@ def get_tokens_from_command_line_flags(  #
...
-        continue
-      tokens.add(line)
+  return sorted(set(model_tokens) | set(tokens_from_file))
+
+
+def get_annotation_names_from_command_line_flags() -> Sequence[str]:
+  """Returns the list of annotation names used in the model.
+
+  When the command-line flag --gematria_annotation_names_file is used, returns a
+  sorted list of annotation_names from this file. The input file is expected to
```
ondrasej, line 171: Very nit: s/annotation\_names/annotation names/

gematria/model/python/token_model_flags.py
```diff
@@ -114,25 +157,40 @@ def get_tokens_from_command_line_flags(  #
...
+  be a text file that contains one annotation name per line. When loading the
+  annotation names, the function:
+    1. removes leading and trailing whitespace from each line.
+    2. ignores lines starting with a hash character (#).
+
+  Returns:
+    The list of annotation_names loaded from the file specified by the
```
ondrasej, line 178: Same here.

gematria/testing/testdata/annotated_basic_blocks_with_throughput.pbtxt
ondrasej, line 1: This is a copy of the original basic_blocks_with_throughput, right?

gematria/granite/python/graph_builder_model_base.py
```diff
@@ -143,6 +160,21 @@ def __init__(
 
     self._special_tokens_tensor = None
 
+    self._instruction_node_mask = tf.placeholder(
+        dtype=tf.dtypes.bool,
+        shape=(None,),
+        name=GraphBuilderModelBase.INSTRUCTION_NODE_MASK_TENSOR_NAME,
+    )
```
ondrasej, line 167: The two `tf.placeholder()`s should be created in `self._create_tf_graph()`, along with the rest of the TF ops.

I remember that in the past creating tensors in the constructor led to different issues with having TF ops in two different TF graphs (which then caused execution issues).

ondrasej (changes requested):

gematria/granite/python/token_graph_builder_model.py
virajbshah: Ah, got it.

Regarding the slicing - the reason why it currently isn't as clean as a simple concatenate operation is that we are only updating instruction nodes, so we'd have to handle the last `num_annotation` entries of non-instruction nodes separately.

A possible solution would be to have two `snt.Embed`s composed in this class - one holding the learnt embeddings of length `embed_dim - num_annotations` for all nodes, and another holding learnt embeddings of length `num_annotations` for only non-instruction nodes, which would be interleaved with `instruction_annotations` accordingly to the `instruction_node_mask` and then concatenated with the other one.

Or instead of having the second set of learnt embeddings we could just use a larger `embed_dim` and let the last `num_annotation` embedding entries for non-instruction nodes be something like zeros - though that sounds like it might be wasteful.

gematria/granite/python/token_graph_builder_model.py
virajbshah: I do remember testing TFLite inference after these changes, and I didn't see anything related to the graphdef having unsupported ops. I can see scatter and gather related ops in the TFLite MLIR dialect `tfl` as well - so it looks like we are safe here :)

gematria/model/python/token_model_flags.py
virajbshah: Ah, I'm wondering how that happened :)

gematria/testing/testdata/annotated_basic_blocks_with_throughput.pbtxt
virajbshah: Yes - just with random instruction annotations added to it.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -378,8 +366,9 @@ def _make_batch_feed_dict(self) -> model_base.FeedDict:
     return feed_dict
 
 
-class TokenGraphBuilderModelNodeEmbed(snt.Embed):
-  """Extends `snt.Embed` to include instruction annotations in node embeddings.
+class TokenGraphBuilderModelNodeEmbed:
+  """`snt.Embed`-like class representing node embeddings with instruction
+  annotations included.
```
ondrasej, line 371: Very nitpicky: the first line of the docstring should fit into 80 columns. I think we can move the note about `snt.Embed` to the extended part of the docstring.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -388,52 +377,64 @@ class TokenGraphBuilderModelNodeEmbed(snt.Embed):
...
-    if self.num_annotations > self.embed_dim:
-      raise ValueError('num_annotations cannot be greater than embed_dim.')
-
     self.instruction_annotations = instruction_annotations
     self.instruction_node_mask = instruction_node_mask
 
-  def __call__(self, inputs):
-    embeddings = super().__call__(inputs)
+    # The first `embed_dim - num_annotations` embedding values for all nodes.
+    self.common_embed = snt.Embed(
```
ondrasej, line 402: We should probably mark all the attributes as internal (by using a leading underscore). I don't see a need for a user to access those.

ondrasej (changes requested):
Last few minor changes, otherwise this looks good.

gematria/granite/python/token_graph_builder_model.py
ondrasej: I don't think we have the export to TFLite on any testing workflow (and I've been surprised by what is _not_ supported in the past). But let's go with this and fix problems when we hit them.

ondrasej (approved):

PR #95, opened 2024-03-26
Add option category to BB processing script

boomanaiden154:
This patch adds an option category to the BB processing script, which separates the options specifically for the script from the other options that end up showing on the list from around LLVM. This significantly improves the readability of the help page, at least when looking for options specifically related to the script (which is essentially all the time).


boomanaiden154:
Note that this is stacked on #90.

ondrasej (approved):

PR #96, opened 2024-03-26
Add option to BB filtering script to report progress

boomanaiden154:
This patch adds a --report-progress-every flag to the BB processing script. This is a nice to have when processing extremely large datasets to report status to give an indication of how long it will take and to indicate that the script isn't hung. This is really nice to have on datasets in the ~1M block range and probably essential in the 100M+ block range.


boomanaiden154:
Note that this is stacked on #95.

ondrasej (approved):

PR #97, opened 2024-03-27
Add BHive Comparison script

boomanaiden154:
This patch adds a script for comparing BHive format CSVs (the first column being the hex bb and the second column being the throughput value). The script just outputs the average deviation between the two values.


boomanaiden154:
Opening this as a draft for now as it needs some additional work and isn't ready for review. Need something to remind me that it exists though when I go back to find it.

gematria/datasets/compare_bhive_csv.cc
ondrasej, line 1: We should have tests for the tool - and ideally also split this into cc\_library + cc\_binary with most code in the library, and unit test the library.

ondrasej (changes requested):

gematria/datasets/compare_bhive_csv.cc
```diff
@@ -0,0 +1,87 @@
...
+      ExitOnErr(llvm::make_error<StringError>(errc::invalid_argument,
+                                              "No comma found on input line"));
+
+    std::string_view ThroughputValue1String =
+        std::string_view(File1Line).substr(Line1CommaIndex + 1,
+                                           File1Line.size());
+
+    std::string_view ThroughputValue2String =
+        std::string_view(File2Line).substr(Line2CommaIndex + 1,
+                                           File2Line.size());
```
ondrasej, line 65: We should either check that the blocks are the same or load both files to memory, and join them by the block hex before comparing. Otherwise, this might produce misleading results :)

boomanaiden154:
Closing. It will be better to do this in python where we can use the tfrecord parsers to also compare against tfrecord datasets.

PR #98, opened 2024-03-28
Add support for models consuming annotations to C++ inference API.

virajbshah:
Updates the model conversion script and the C++ inference API to supply instruction annotations to tflite models.


gematria/granite/convert_gb_token_model_to_tflite.sh
```diff
@@ -37,6 +37,8 @@ function print_error_and_exit() {
 # Parse command-line flags.
 # TODO(ondrasej): Consider using getopt instead of parsing the flags manually.
 gematria_export_as_seq2seq=0
+gematria_export_with_deltas=0
```
ondrasej, line 40: IIRC, seq2seq and deltas are the same things under two slightly different names (which is, sadly, confusing).

The way the model is designed, the graph builder models should just always be seq2seq (because the non-seq2seq version doesn't work great).

gematria/granite/convert_gb_token_model_to_tflite.sh
```diff
@@ -78,19 +86,23 @@ function str_join() {
...
   GnnModelBase.senders
   GnnModelBase.num_edges
   GnnModelBase.num_nodes
-  GraphBuilderModelBase.instruction_node_mask
+  $([[ "${gematria_export_with_annotations}" -eq 1 ]] && \
+    echo 'GraphBuilderModelBase.instruction_node_mask')
+  $([[ "${gematria_export_with_deltas}" -eq 1 || \
+       "${gematria_export_with_annotations}" -eq 1 ]] && \
+    echo 'TokenGraphBuilderModel.instruction_annotations')
 )
```
ondrasej, line 106: A more readable version would do something like
```bash
declare -a INPUT_TENSORS_LIST=()
if ...; then
  INPUT_TENSORS_LIST+=(...)
fi
INPUT_TENSORS_LIST+=(...)
...
readonly INPUT_TENSORS_LIST
```

This way, the list is read-only in the end, but we can use simpler way to compose it. Also, ideally, we should use just tensor names and do the index matching from there, but that's for another series of patches.

gematria/granite/convert_gb_token_model_to_tflite.sh
```diff
@@ -103,20 +115,16 @@ readonly TARGET_OPS_LIST=(
...
+readonly OUTPUT_TENSOR_LIST=(
+  ModelBase.output_tensor
+  $([[ "${gematria_export_as_seq2seq}" -eq 1 || \
+       "${gematria_export_with_deltas}" -eq 1 ]] && \
+    echo 'ModelBase.output_tensor_deltas')
+  TokenModel.token_list
+  GraphBuilderModelBase.special_tokens
+  $([[ "${gematria_export_with_annotations}" -eq 1 ]] && \
+    echo 'TokenGraphBuilderModel.annotation_names')
+)
```
ondrasej, line 127: Same here.

ondrasej (commented):
Submitting first few comments. I still need to read through most of the inference .cc file.

gematria/granite/convert_gb_token_model_to_tflite.sh
virajbshah: I initially chose this (admittedly strange) way of doing this as a concise way of letting optional tensors be at any position since IIRC the ordering of tensors needed to match that of the tflite model (though this PR also includes name-based matching, which should eliminate that requirement).

gematria/granite/convert_gb_token_model_to_tflite.sh
virajbshah: Got it. I'll combine the two flags (but keep support for seq2num around, i.e. keep seq2seq behind a flag as opposed to it being the only option).

gematria/granite/convert_gb_token_model_to_tflite.sh
ondrasej: The += syntax would still allow us to add at any spot in the list (and we'd avoid the inline syntax with echos):
```sh
declare -a INPUT_TENSORS_LIST=()
if (( "${gematria_export_as_seq2seq}" )); then
  INPUT_TENSORS_LIST+=(ModelBase.delta_block_index_tensor)
fi
INPUT_TENSORS_LIST+=(
  GnnModelBase.node_features
  GnnModelBase.edge_features
  GnnModelBase.global_features
  GnnModelBase.receivers
  GnnModelBase.senders
  GnnModelBase.num_edges
  GnnModelBase.num_nodes
)
if (( "${gematria_export_with_annotations}" )); then
  INPUT_TENSORS+=(GraphBuilderModelBase.instruction_node_mask)
fi
if (( "${gematria_export_with_deltas}"
      && "${gematria_export_with_annotations}" )); then
  INPUT_TENSORS+=(TokenGraphBuilderModel.instruction_annotations)
fi
...
readonly INPUT_TENSORS_LIST
```

Same for the output list.

gematria/granite/graph_builder_model_inference.cc
ondrasej, line 1: The move from positional inputs/outputs to named inputs/outputs is independent from the changes we need to do for annotations. If possible (if it's not too much work), I'd prefer splitting it out into a separate PR, so that each change has a limited surface.

ondrasej (changes requested):

boomanaiden154:
I'm assuming this patch needs to land for the `CMake` build to be fixed?

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -149,7 +153,7 @@ constexpr InputTensor kRequiredInputTensors[]{
     InputTensor::kGraphNodes,   InputTensor::kGraphEdges,
     InputTensor::kGraphGlobals, InputTensor::kGraphReceivers,
     InputTensor::kGraphSenders, InputTensor::kGraphNEdge,
-    InputTensor::kGraphNNode,   InputTensor::kInstructionNodeMask,
```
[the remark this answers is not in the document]
virajbshah: IIRC, the instruction node mask never gets used by `seq2num` models without annotated instructions.

Just to be sure, I just checked the exported `GraphDef` for such a model, and `instruction_node_mask` indeed gets pruned out 🙂 

There are also checks for it when it is needed, so it should be all right this way even though we probably won't exactly be training too many `seq2num` models.

gematria/granite/python/token_graph_builder_model.py
```diff
@@ -22,9 +22,11 @@
 from gematria.granite.python import graph_builder
 from gematria.granite.python import graph_builder_model_base
 from gematria.model.python import model_blocks
+from gematria.model.python import model_base
 from gematria.model.python import options
 import graph_nets
 import sonnet as snt
+import numpy as np
 import tensorflow.compat.v1 as tf
```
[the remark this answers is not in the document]
virajbshah: Oops, those slipped through from the merge. They were needed for something that was both added and removed in this stack of PRs 🙂

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -433,6 +439,39 @@ llvm::Expected<std::string> GetNodeTokenAtIndex(
...
+    return llvm::createStringError(llvm::errc::invalid_argument,
+                                   "The annotation names could not be read");
+  }
+  const std::string_view annotation_names_data(annotation_names_raw_data,
+                                               annotation_names_size_bytes);
+  std::vector<std::string> annotation_names =
+      StrSplitAsCopy(annotation_names_data, '\0');
+  return std::vector<std::string>(
+      std::make_move_iterator(annotation_names.begin()),
+      std::make_move_iterator(annotation_names.end()));
```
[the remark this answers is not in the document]
virajbshah: Right, I'm guessing that is artifact from when the annotation names were held by a set, and the set had to be built from the vector.

virajbshah:
I haven't investigated what exactly it is, but interestingly the `check-llvm-tools-llvm-cm` failure is a result of some relatively recent change to `llvm`.

ondrasej (approved):

ondrasej:
I relaunched the CMake CI job, let's see if this helps. If not, you might need to merge or rebase from `main`.

PR #99, opened 2024-03-29
Add a requirements lockfile

boomanaiden154:
This patch adds a requirements lockfile. This enables reproducible python dependency installation and also prevents issues cropping up from new versions being released. This patch defaults to `tensorflow==2.13.1` due to some issues that cropped when bumping past that. These should be resolved, but can be done so in a future patch.


boomanaiden154:
This fixed some issues I was having, and as a bonus, fixes the CI too.

ondrasej (approved):

PR #100, opened 2024-04-01
Skip syscall instructions in the process bb script

boomanaiden154:
This patch enables skipping system call instructions in the BB processing script. Depending upon the system call instruction, it might not just return and can potentially hang the block.


gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -79,6 +82,7 @@ Expected<std::string> ProcessBasicBlock(
        *DisassembledInstructionsOrErr) {
     MCInstrDesc InstDesc =
         LLVMSupport.mc_instr_info().get(Instruction.mc_inst.getOpcode());
+    if (Instruction.mc_inst.getOpcode() == X86::SYSCALL) continue;
```
ondrasej, line 85: There is a whole category of instructions that may have side-effects that are not modelled by the other fields of the descriptor. We probably want to skip all of those.

See https://github.com/llvm-mirror/llvm/blob/2c4ca6832fa6b306ee6a7010bfb80a3f2596f824/include/llvm/MC/MCInstrDesc.h#L443 for more details.

ondrasej (approved):

gematria/datasets/process_and_filter_bbs.cc
boomanaiden154: Good point. I didn't realize that was a field of `MCInstDesc`. I've added a todo for now as I know filtering syscalls works to reasonable effectiveness and am not sure what the consequences of filtering everything with non-modelled side effects is at scale.

PR #101, opened 2024-04-02
Parallelize BB processing script

boomanaiden154:
This patch parallelizes the BB processing script. This significantly speeds up the processing of BBs. Eventually diminishing returns are reached, especially on systems with a large number of threads, most likely due to blocking on IO.


boomanaiden154:
I'm seeing near linear speedup up to about 16 threads and then I start to hit diminishing returns.

boomanaiden154:
Converting to a draft as it needs some more work. Currently not handling the case where a batch is left with some items at the end.

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,166 @@
...
+    cl::init(""), cl::cat(ProcessFilterCat));
+
+static cl::opt<std::string> OutputFile(
+    "output-file",
+    cl::desc(
+        "Path to the output CSV file with processed/filtered basic blocks"),
+    cl::init(""), cl::cat(ProcessFilterCat));
+
+static cl::opt<bool> FilterMemoryAccessingBlocks(
+    "filter-memory-accessing-blocks",
```
ondrasej, line 48: I'd change this to `remove-memory-accessing-blocks` to make it immediately obvious that memory-accessing blocks are removed, not the only ones preserved.

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,166 @@
...
+int main(int Argc, char **Argv) {
+  cl::ParseCommandLineOptions(Argc, Argv, "process_and_filter_bbs");
+
+  ExitOnError ExitOnErr("process_and_filter_bbs error: ");
+
+  const std::unique_ptr<gematria::LlvmArchitectureSupport> LLVMSupport =
+      gematria::LlvmArchitectureSupport::X86_64();
+
+  std::unique_ptr<MCInstPrinter> MachineInstructionPrinter =
+      LLVMSupport->CreateMCInstPrinter(0);
```
ondrasej, line 119: When we run multiple threads, is it safe to use the same instruction printer from all of them at the same time?

gematria/datasets/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,166 @@
...
+  for (const gematria::DisassembledInstruction &Instruction :
+       *DisassembledInstructionsOrErr) {
+    MCInstrDesc InstDesc =
+        LLVMSupport.mc_instr_info().get(Instruction.mc_inst.getOpcode());
+    if (Instruction.mc_inst.getOpcode() == X86::SYSCALL) continue;
+    if (InstDesc.isReturn() || InstDesc.isCall() || InstDesc.isBranch())
+      continue;
+    if (FilterMemoryAccessingBlocks &&
+        (InstDesc.mayLoad() || InstDesc.mayStore()))
+      continue;
```
ondrasej, line 103: IIUC, this will remove control-affecting instructions out of the block, but it preserves the rest of the instructions. I understand that/why we want to do this for branches and other control-affecting instructions, but:
1. we should check that they are at the end of the block. If they are elsewhere, we're distorting the block to something that is not realistic, and probably not correct.
2. IIUC, for blocks accessing memory, we want to drop the whole block, not just the memory-accessing instructions, which is what we do here.

ondrasej (changes requested):

boomanaiden154:
Closing this for now. I think I want to restructure this to do something different rather than just parallelizing in process. Adding Python bindings and using something like Apache Beam I think would enable more scalability for dataset processing.

PR #102, opened 2024-04-04
Add a script to produce vocab from a dataset

boomanaiden154:
This patch adds in a script to produce a vocab file from a dataset that can then be used to train ML models requiring a list of vocab words to generate an embeddings table.


gematria/datasets/python/generate_vocab.py
```diff
@@ -0,0 +1,73 @@
...
+    None,
+    'The path to the tfrecord file to process',
+    required=True,
+)
+_OUTPUT_TXT_FILE = flags.DEFINE_string(
+    'output_vocab', None, 'The path to the output txt file', required=True
+)
+
+
+def main(_) -> None:
```
ondrasej, line 47: Nit (to keep in line with the Google style guide): `_` for unused args is discouraged, because the name of the arg is part of the function signature (for keyword arg calls), and changing it might break code.

So, to avoid having to change the name, the recommendation is to use the name as usual (here, `argv`) and then `del` it in the body of the function to mark it as unused.

```
del argv  # Unused.
```
(the style guide requires that `del` always comes with a comment)

gematria/datasets/python/generate_vocab.py
```diff
@@ -0,0 +1,73 @@
...
+    basic_block_proto = basic_block_protos.basic_block_from_proto(
+        proto.basic_block
+    )
+    for instruction in basic_block_proto.instructions:
+      tokens.update(instruction.as_token_list())
+
+    current_proto_index += 1
+
+  with open(_OUTPUT_TXT_FILE.value, 'w') as output_file_handle:
+    for token in tokens:
```
ondrasej, line 68: I'd do `sorted(tokens)` here. It makes the token file easier to work with.

ondrasej (changes requested):
It would be good to have a small (LIT) test for the tool.

boomanaiden154:
Addressed feedback, added a test.

gematria/datasets/python/generate_vocab.py
```diff
@@ -0,0 +1,75 @@
...
+from gematria.basic_block.python import basic_block
+
+_INPUT_TFRECORD_FILE = flags.DEFINE_string(
+    'input_tfrecord',
+    None,
+    'The path to the tfrecord file to process',
+    required=True,
+)
+_OUTPUT_TXT_FILE = flags.DEFINE_string(
+    'output_vocab', None, 'The path to the output txt file', required=True
```
ondrasej, line 43: `--gematria_output_tokens_file` to be consistent with the flags used by the models.

gematria/datasets/python/generate_vocab.py
```diff
@@ -0,0 +1,75 @@
...
+from absl import flags
+from absl import logging
+
+from gematria.io.python import tfrecord
+from gematria.proto import throughput_pb2
+from gematria.basic_block.python import basic_block_protos
+from gematria.basic_block.python import basic_block
+
+_INPUT_TFRECORD_FILE = flags.DEFINE_string(
+    'input_tfrecord',
```
ondrasej, line 37: `--gematria_input_tfrecord`. Command-line flags are global variables, so it's usually safer to use prefixes...

gematria/datasets/python/generate_vocab.py
```diff
@@ -0,0 +1,75 @@
...
+
+  loaded_protos = tfrecord.read_protos(
+      _INPUT_TFRECORD_FILE.value, throughput_pb2.BasicBlockWithThroughputProto
+  )
+
+  tokens = set()
+
+  current_proto_index = 0
+
+  for proto in loaded_protos:
```
ondrasej, line 58: More idiomatic Python:

```python
for current_proto_index, proto in enumerate(loaded_protos):
  ...
```

ondrasej (changes requested):
Thanks for working on this!

ondrasej (approved):

PR #103, opened 2024-04-05
Switch to using fast cpp protos

boomanaiden154:
This patch sets --define=use_fast_cpp_protos=true to ensure that we are using the cpp protos instead of the python protos. This results in about an 85% reduction in time spent loading protos due to the lack of needing to serialize/deserialize at the language interface.


ondrasej (approved):
Thanks for cleaning this up!

PR #104, opened 2024-04-06
Add support for filtering eflags operands

boomanaiden154:


boomanaiden154:
Not sure why I thought this was a good idea.

PR #119, opened 2024-06-04
FindAccessedAddrs support for finding accessed registers

orodley:


ondrasej (changes requested):

gematria/datasets/find_accessed_addrs_from_bhive.cc
```diff
@@ -34,6 +39,47 @@ ABSL_FLAG(bool, quiet, false, "Omit all output except for the final summary");
...
+  if (!value) return;
+  snippets_file << "# LLVM-EXEGESIS-DEFREG " << name << " "
+                << absl::StrFormat("%016x", *value) << "\n";
+}
+
+void WriteExegesisSnippet(gematria::BHiveImporter& bhive_importer,
+                          std::string_view snippets_dir,
+                          const std::vector<uint8_t>& code,
+                          const gematria::AccessedAddrs& addrs, int n) {
+  auto proto = bhive_importer.BasicBlockProtoFromMachineCode(code);
```
ondrasej, line 56: One more TODO to add (feel free to assign to me or Aiden): merge the support code in both versions of the accessed addresses code.

ondrasej (approved):

gematria/datasets/find_accessed_addrs_from_bhive.cc
boomanaiden154: It seems like `convert_bhive_to_llvm_exegesis_input.cc` is pretty similar to this script? I think we can do more than just merge the support code and probably just deprecate one or the other.

gematria/datasets/find_accessed_addrs.cc
```diff
@@ -410,35 +459,408 @@ absl::Status ForkAndTestAddresses(absl::Span<const uint8_t> basic_block,
...
-  regs.r9 = random_reg();
-  regs.r10 = random_reg();
-  regs.r11 = random_reg();
-  regs.r12 = random_reg();
-  regs.r13 = random_reg();
-  regs.r14 = random_reg();
-  regs.r15 = random_reg();
+std::optional<int64_t>* LLVMRegNumberToX64Reg(X64Regs& regs,
+                                              int llvm_reg_number) {
+  switch (llvm_reg_number) {
```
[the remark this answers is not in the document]
orodley: Done.

gematria/datasets/find_accessed_addrs.cc
```diff
@@ -410,35 +459,408 @@ absl::Status ForkAndTestAddresses(absl::Span<const uint8_t> basic_block,
...
+    default:
+      return nullptr;
+  }
 }
 
 }  // namespace
 
+X64Regs FindReadRegs(const LlvmArchitectureSupport& llvm_arch_support,
+                     absl::Span<const uint8_t> basic_block) {
+  X64Regs regs;
```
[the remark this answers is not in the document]
orodley: Done.

gematria/datasets/find_accessed_addrs_from_bhive.cc
orodley: Thanks for pointing me to that file. Yeah there's enough overlap that I think you're right. Looks like `convert_bhive_to_llvm_exegesis_input.cc` already emits these directives, but just uses hardcoded initial register values and memory values. Adding a flag to that tool to instead use `FindAccessedAddrs` to determine these values would make more sense than duplicating that here.

I initially created this tool just to have a way to test `FindAccessedAddrs` on a large dataset. If we do the above change then `convert_bhive_to_llvm_exegesis_input.cc` will serve that purpose perfectly well and `find_accessed_addrs_from_bhive.cc` can be removed. I'll add the TODO.

PR #125, opened 2024-06-22
Add sanitizer configs to bazel

boomanaiden154:
This patch adds sanitizer configs to .bazelrc so a developer can simply use --config=asan or --config=tsan on a specific test to see if there are any issues. It seems like asan might be catching some issues I will have to look into this in a bit.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -8,4 +8,18 @@ build \
   --cxxopt=-std=c++17 \
   --host_cxxopt=-std=c++17 \
   --repo_env=CC=clang \
-  --define=use_fast_cpp_protos=true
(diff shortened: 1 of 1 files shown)


ondrasej (approved):
LGTM in general.

That said, they are usually not very useful when dealing with inline assembly, JITs and other low-level features.

boomanaiden154:
> That said, they are usually not very useful when dealing with inline assembly, JITs and other low-level features.

Right. This was mostly intended for looking at dataset processing things that don't have to do anything low-level (i.e., not benchmarking this case). The motivating example as #101 with the question of whether or not we need a per-thread `MCInstPrinter`.

PR #127, opened 2024-06-22
Bump clang format version in CI to v18

boomanaiden154:
This patch bumps the clang format version in CI to v18 as v18 is the latest LLVM release (currently) and includes quite a few patches for clang-format.


boomanaiden154:
This will be `clang-format` clean when #126 lands.

ondrasej (approved):
Thanks for keeping this up to date!

PR #129, opened 2024-06-26
Make llvm-cm tests invariant against small FP differences

boomanaiden154:
Currently the llvm-cm tests are quite sensitive to small FP differences as they evaluate a model where the results are not guaranteed to be exact across hardware. Check the two most significant figures and then regex that we have numbers to make them invariant against slight changes.


ondrasej (approved):
👍

PR #131, opened 2024-06-29
Refactor find_accessed_addrs to use BasicBlockUtils

boomanaiden154:
This patch refactors FindAccessedAddrs to use the BasicBlockUtils library for getting the registers used in a block. This patch also does some other refactorings to make the format used in find_accessed_addrs more consistent with what is used elsewhere in the code base, particularly changing the used register set to a vector of registers/values.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -99,13 +99,14 @@ ABSL_FLAG(unsigned, max_annotation_attempts, 50,
 absl::StatusOr<gematria::AccessedAddrs> GetAccessedAddrs(
     absl::Span<const uint8_t> basic_block,
     gematria::ExegesisAnnotator* exegesis_annotator,
-    const unsigned max_annotation_attempts) {
+    const unsigned max_annotation_attempts,
+    gematria::LlvmArchitectureSupport& llvm_support) {
   const AnnotatorType annotator_implementation =
       absl::GetFlag(FLAGS_annotator_implementation);
   switch (annotator_implementation) {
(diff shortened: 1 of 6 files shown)


gematria/datasets/find_accessed_addrs.h
```diff
@@ -25,103 +25,45 @@
...
-      absl::FunctionRef<void(typename Templ<int64_t>::type&)> visitor) {
-    ForEachReg([visitor](typename Templ<int64_t>::type& reg,
-                         std::string_view name) { visitor(reg); });
-  }
-};
-
-template <typename T>
-struct Optional {
-  using type = std::optional<T>;
+struct RawX64Regs {
```
orodley, line 29: Can you add a comment on this struct describing its purpose? Something based on the comment previously at line 102 would be fine.

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -99,13 +99,14 @@ ABSL_FLAG(unsigned, max_annotation_attempts, 50,
...
     absl::Span<const uint8_t> basic_block,
     gematria::ExegesisAnnotator* exegesis_annotator,
-    const unsigned max_annotation_attempts) {
+    const unsigned max_annotation_attempts,
+    gematria::LlvmArchitectureSupport& llvm_support) {
   const AnnotatorType annotator_implementation =
       absl::GetFlag(FLAGS_annotator_implementation);
   switch (annotator_implementation) {
     case AnnotatorType::kFast:
       // This will only get the first segfault address.
```
orodley, line 108: Side note: this comment is no longer true

orodley (approved):
Looks good, thanks!

PR #132, opened 2024-07-03
Apply fix to keep using Keras 2

virajbshah:
* TensorFlow 2.16 uses Keras 3 by default, which is incompatible with
   `tensorflow-probability` as of version 0.24.0.
 * We use `tf_keras` 2.16 instead of `tensorflow.keras` to keep using
   Keras 2 without breakages.


boomanaiden154:
Just noting in case you haven't seen it yet, but you'll also need to regenerate the lockfile (`requirements.txt`) from `requirements.in` so that the CI will pick up the new dependency.

virajbshah:
Yep, just noticed that while looking at the CI logs. There's also some discrepancy with the Python formatting check (maybe because of using `--config=pyproject.toml`).

virajbshah:
On a second look, the Python formatting difference seems to come from the `pyink` version, not the `config`, which is a little strange.

boomanaiden154:
> On a second look, the Python formatting difference seems to come from the pyink version, not the config, which is a little strange.

Python formatting can be really sensitive to the formatter version. That's why the version is pinned in the CI. We should probably stick the CI/development deps in a lockfile somewhere to alleviate issues like that (`requirements-dev.txt`?), but I haven't had a major reason to do it yet.

gematria/granite/python/gnn_model_base.py
```diff
@@ -27,6 +27,7 @@
 import graph_nets
 import sonnet as snt
 import tensorflow.compat.v1 as tf
+import tf_keras as keras
```
ondrasej, line 30: Do we need the module rename? This doesn't seem to be a standard rename, so for the sake clarity, I'd just use `tf_keras` in the code.

Same in the other files.

gematria/granite/python/gnn_model_base_test.py
```diff
@@ -457,8 +458,8 @@ def test_train_seq2num_encoder_decoder_model(
         learning_rate=0.01,
     )
     with mock.patch(
-        'tensorflow.compat.v1.keras.layers.LayerNormalization',
-        side_effect=tf.keras.layers.LayerNormalization,
+        'tf_keras.layers.LayerNormalization',
+        side_effect=keras.layers.LayerNormalization,
```
ondrasej, line 462: If this makes the tests pass, then no change needed in this PR. But I'm wondering if we need this patch at all, and maybe we could remove it altogether?

ondrasej:
> > On a second look, the Python formatting difference seems to come from the pyink version, not the config, which is a little strange.
> 
> Python formatting can be really sensitive to the formatter version. That's why the version is pinned in the CI. We should probably stick the CI/development deps in a lockfile somewhere to alleviate issues like that (`requirements-dev.txt`?), but I haven't had a major reason to do it yet.

Maybe we should just go with the flow and reformat from time to time. In the long run, this might be less painful than trying to stick with a pinned version and having to deal with formatting each time we update the internal version from GitHub.

ondrasej (approved):
Cool, thanks for working on this!

boomanaiden154:
> Maybe we should just go with the flow and reformat from time to time. In the long run, this might be less painful than trying to stick with a pinned version and having to deal with formatting each time we update the internal version from GitHub.

Seems reasonable enough to me. We still probably need to have a pinned version somewhere though and point contributors to that because otherwise you get a potential formatting difference for every new formatter release that hasn't yet been "ingested" by us, which might take a while since it just happens whenever someone gets around to it.

gematria/granite/python/gnn_model_base.py
virajbshah: It's not really needed, but is just what the release notes mentioning the fix did, presumably so that you would only really need to change the import line if you originally had something like `from tensorflow import keras` (which was not our case).

ondrasej (approved):
LGTM for the renaming. And as we discussed in person, let's investigate and maybe remove the patching of LayerNorm in another PR.

PR #134, opened 2024-07-14
Use information from AccessedAddrs in conversion tool

boomanaiden154:
This patch makes convert_bhive_to_llvm_exegesis_inputs use information from AccessedAddrs like the initial register values and the block contents to emit information to the file rather than assuming a default register value/default block content.

### DIFF
--- a/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
+++ b/gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
@@ -39,12 +39,6 @@
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/tools/llvm-exegesis/lib/TargetSelect.h"
 
-// Use the constants from the BHive paper for setting initial register and
-// memory values. These constants are set to a high enough value to avoid
-// underflow and accesses within the first page, but low enough to avoid
-// exceeding the virtual address space ceiling in most cases.
-constexpr uint64_t kInitialRegVal = 0x12345600;
-constexpr uint64_t kInitialMemVal = 0x12345600;
 constexpr unsigned kInitialMemValBitWidth = 64;
 constexpr std::string_view kRegDefPrefix = "# LLVM-EXEGESIS-DEFREG ";
 constexpr std::string_view kMemDefPrefix = "# LLVM-EXEGESIS-MEM-DEF ";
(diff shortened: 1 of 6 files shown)


gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
```diff
@@ -213,17 +201,30 @@ absl::Status WriteAsmOutput(const AnnotatedBlock& annotated_block,
...
+    std::string register_value_string =
+        gematria::ConvertHexToString(register_and_value.register_value);
+    output_file << kRegDefPrefix
+                << reg_info.getName(register_and_value.register_index) << " "
+                << register_value_string << "\n";
   }
 
   // Multiple mappings can point to the same definition.
   if (annotated_block.accessed_addrs.accessed_blocks.size() > 0) {
+    std::string memory_value_string = gematria::ConvertHexToString(
```
orodley, line 215: (I see this code was just moved around, not newly written, so feel free to ignore)

Consider using one of the Abseil zero-padding options instead of doing it manually. Something like:

`absl::StrCat(absl::Hex(annotated_block.accessed_addrs.block_contents, absl::kZeroPad16))`

or

`absl::StrFormat("%016x", annotated_block.accessed_addrs.block_contents)`

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -183,6 +191,16 @@ Expected<AccessedAddrs> ExegesisAnnotator::findAccessedAddrs(
     MemAnnotations.accessed_blocks.push_back(Mapping.Address);
   }
 
+  std::vector<unsigned> UsedRegisters = gematria::getUsedRegisters(
+      *DisInstructions, State.getRegInfo(), State.getInstrInfo());
+
```
orodley, line 196: Nit: `MemAnnotations.initial_regs.reserve(UsedRegisters.size())`

orodley (approved):

gematria/datasets/convert_bhive_to_llvm_exegesis_input.cc
boomanaiden154: Good point. Updated to use `absl::StrFormat`.

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Done.

PR #138, opened 2024-07-17
Add `annotating_importer` utility.

virajbshah:
A library that can:
 * Extract basic blocks from an ELF file.
 * Extract samples and LBR data from a `perf.data`-like file.
 * Use them to create annotated Gematria protos.


gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+  std::unordered_map<uint64_t, std::vector<int>> samples;
+  for (const auto &event : perf_data_proto.events()) {
+    // Filter out non-sample events.
+    if (!event.has_sample_event()) {
+      continue;
+    }
+
+    // Filter out sample events from outside the profiled binary.
+    uint64_t sample_ip = event.sample_event().ip();
+    if (!(mmap.start() <= sample_ip && sample_ip < mmap.start() + mmap.len())) {
```
ondrasej, line 205: 1. Simpler, with De Morgan's laws:
```
if (sample_ip < mmap_start || sample_id >= mmap_end) continue;
```

2. `mmap.start() + mmap.end()` is always the same, since we track just one `mmap` event, so we could pre-compute it for shorter and more efficient code.

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+
+    if (!samples.count(sample_ip)) {
+      samples[sample_ip] = std::vector<int>(sample_types.size(), 0);
+    }
+    int event_idx = 0;
+    if (has_multiple_sample_types) {
+      event_idx =
+          event_code_to_idx[event_id_to_code[event.sample_event().id()]];
+    }
+    samples[sample_ip][event_idx] += 1;
```
ondrasej, line 217: Simpler version that avoids multiple hashmap lookups:
```
std::vector<int>& samples_for_ip = samples[sample_ip];
if (samples_for_ip.empty()) samples_for_ip.resize(sample_types.size());
int event_idx = 0;
if (has_multiple_sample_types) { /* ... */ }
samples_for_ip[event_idx] += 1;
```

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    // Filter out sample events from outside the profiled binary.
+    uint64_t sample_ip = event.sample_event().ip();
+    if (!(mmap.start() <= sample_ip && sample_ip < mmap.start() + mmap.len())) {
+      continue;
+    }
+
+    if (!samples.count(sample_ip)) {
+      samples[sample_ip] = std::vector<int>(sample_types.size(), 0);
```
ondrasej, line 210: I'd prefer defining
```
const int num_sample_types;
```
somewhere at the top of the function, and using that throughout. In some cases (if alias analysis can't prove that it never changes), `sample_types.size()` might require memory access.

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+  if (!perf_parser_.ParseRawEvents()) {
+    return absl::InvalidArgumentError(
+        "The given `perf.data`-like file could not be parsed.");
+  }
+
+  return absl::OkStatus();
+}
+
+absl::StatusOr<std::vector<
+    std::tuple<uint64_t, uint64_t, std::vector<DisassembledInstruction>>>>
```
ondrasej, line 70: As a general comment: most of the time, I'd prefer to avoid `std::pair` and `std::tuple` and use a struct instead. It often produces more readable code that needs less documentation.

In this specific case: do we need the start + end address? When `DisassembledInstruction` is filled in completely, then `start == instructions.front().address` and `end == instructions.back().address + instructions.back().machine_code.size()`, no?

gematria/datasets/annotating_importer.h
```diff
@@ -0,0 +1,90 @@
...
+  // a mapping between sample addresses and the corresponding sample values.
+  // The ordering of the sample values matches the ordering of types in the
+  // heading.
+  absl::StatusOr<std::pair<std::vector<std::string>,
+                           std::unordered_map<uint64_t, std::vector<int>>>>
+  GetSamples();
+
+  // Extracts start and end pairs, as well as lengths in cycles, of sequences of
+  // straight-run code from branch stacks. Returns an unordered_map:
+  // `start_addr` -> {`end_addr`, [`cycles_measurements`, ...]}.
```
ondrasej, line 67: In the presence of conditional branches, there might be multiple `end_addr`s for single `start_addr`, no? So, we should either have a double map, or use `{start_addr, end_addr}` as the key in the map.

This is based on my understanding that `start_addr` and `end_addr` are the start and the end of the (pseudo-)basic block, i.e. `start_addr` is the destination address of the branch instruction that "entered" the basic block, and `end_addr` is the address of the taken branch instruction that moved to another basic block.

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+      lbr_data;
+  for (const auto &event : perf_data_proto.events()) {
+    if (!event.has_sample_event() ||
+        !event.sample_event().branch_stack_size()) {
+      continue;
+    }
+    const auto &branch_stack = event.sample_event().branch_stack();
+    for (int branch_idx = 0; branch_idx + 1 < branch_stack.size();
+         ++branch_idx) {
+      const auto &next_branch_entry = branch_stack.at(branch_idx);
```
ondrasej, line 238: `branch_stack[branch_idx]` doesn't work?

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+      const auto &branch_entry = branch_stack.at(branch_idx + 1);
+      if (lbr_data.count(branch_entry.to_ip())) {
+        auto &[run_end, cycles_values] = lbr_data[branch_entry.to_ip()];
+        if (next_branch_entry.from_ip() == run_end) {
+          cycles_values.push_back(next_branch_entry.cycles());
+        } else if (next_branch_entry.from_ip() < run_end) {
+          run_end = next_branch_entry.from_ip();
+          cycles_values = std::vector<uint32_t>{next_branch_entry.cycles()};
+        }
```
ondrasej, line 247: As noted in the .h file: we could also preserve all the versions as separate entries. They will have some not taken conditional branches, but that makes them quite interesting for the training.

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+    double throughput_value = -1;
+    if (lbr_data.value().count(block_begin_addr)) {
+      const auto &[run_end_addr, cycles_values] =
+          lbr_data.value().at(block_begin_addr);
+      if (block_end_addr == run_end_addr) {
+        throughput_value =
+            std::accumulate(cycles_values.begin(), cycles_values.end(), 0.0) /
+            cycles_values.size();
+      }
+    }
```
ondrasej, line 333: We might want to take the LBR (pseudo-)blocks. By going through BBAddrMap, we might miss a large number of blocks where we have LBR data, but they do not correspond to a BBAddrMap entry.

This doesn't concern just traces with a not-taken conditional branch, but also branchless instruction sequences that can be entered at multiple addresses:
```
L1:
  inst1
  inst2
L2:
  inst3
  inst4
L3:
  inst5
  inst6
  JMP Lxyz
```
will appear in BBAddrMap as L1-L2, L2-L3, L3-JMP, but we would get L1-JMP, L2-JMP, L3-JMP from LBR and we'd throw out L1-JMP, L2-JMP.

ondrasej (changes requested):
Cool!

WORKSPACE
```diff
@@ -79,6 +79,35 @@ git_repository(
...
 
+# We use `perf_data_converter` with a patch to make it more usable as a library.
+git_repository(
+    name = "com_google_perf_data_converter",
+    commit = "442981cd4071fa9b1057b2609406db027e6d6263",
+    patch_args = ["-p1"],
+    patch_cmds = [
+        "mv ./src/quipper .",
+        "rm -rf ./src",
+    ],
```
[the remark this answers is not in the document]
virajbshah: It does the same thing, but using `patch_cmds` is a bit more convenient since the commands are applied after the patch is, which makes it easier to generate the patch. On the other hand, IIUC, `strip_prefix` is applied before the patch so the diff needs to be from after the `mv` rather than from a fresh clone.

WORKSPACE
```diff
@@ -79,6 +79,35 @@ git_repository(
...
+    strip_prefix = "elfutils-0.191",
+    urls = ["https://sourceware.org/pub/elfutils/0.191/elfutils-0.191.tar.bz2"],
+)
+
+http_archive(
+    name = "boringssl",
+    sha256 = "0a2b7a10fdce3d5ccdc6abf4f5701dca24b97efa75b00d203c50221269605476",
+    strip_prefix = "boringssl-ea4425fbb276871cfec5c4e19c12796b3cd1c9ab",
+    urls = ["https://github.com/google/boringssl/archive/ea4425fbb276871cfec5c4e19c12796b3cd1c9ab.tar.gz"],
+)
```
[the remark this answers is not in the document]
virajbshah: Yep, it is a dependency of `quipper`.

gematria/datasets/annotating_importer.cc
virajbshah: Yep, I realised this and had added a little TODO for myself just beneath on line 140 🙂 

```cpp
// TODO(vbshah): Consider removing the addresses and just using the
// `addr` fields on the instructions.
basic_blocks.emplace_back(begin_idx - file_offset,
                          end_idx - file_offset, instructions.value());
```

gematria/datasets/annotating_importer.h
virajbshah: Sounds right. Initially I wrote this only to try to get basic block latencies so we'd only care about one start-end pair: the one we think corresponds to a basic block. But since it looks like pseudo-blocks will be relevant as well, we should move to using a double map or the different keying.

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+    return absl::InvalidArgumentError(
+        "The given `perf.data`-like file could not be parsed.");
+  }
+
+  return absl::OkStatus();
+}
+
+absl::StatusOr<std::vector<
+    std::tuple<uint64_t, uint64_t, std::vector<DisassembledInstruction>>>>
+AnnotatingImporter::GetBlocksFromELF(std::string_view file_name) {
```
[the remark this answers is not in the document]
virajbshah: Sounds good. We should probably have a library doing all the individual conversions between block representations and everything can take whatever it needs from it:
 * `Binary -> std::string -> DisassembledInstruction -> BasicBlockWithThroughputProto`

gematria/datasets/annotating_importer.cc
```diff
@@ -0,0 +1,340 @@
...
+    // binary always correspond to the first MMapEvent. Implement BuildID or
+    // name based checking.
+    if (event.has_mmap_event() &&
+        event.mmap_event().prot() & 1 /* PROT_READ */ &&
+        event.mmap_event().prot() & 4 /* PROT_EXEC */) {
+      mmap = event.mmap_event();
+      break;
+    }
+  }
+  // TODO(virajbshah): Make sure the mapping was found. (Use num_mmap_events)
```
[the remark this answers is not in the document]
virajbshah: As of now, `quipper` logs out things like the number of MMAP events found and some measures of how many samples could be mapped. I'll sort out this TODO in this PR.

gematria/datasets/annotating_importer.h
```diff
@@ -0,0 +1,120 @@
...
+  // Loads a `perf.data`-like file into the importer. Must be called before
+  // `GetSamples`, `GetLBRData`, and `GetLBRBlocksWithLatency`.
+  absl::Status LoadPerfData(const std::string_view file_name);
+
+  // Loads a binary into the importer for further processing. Must be called
+  // before `GetElfFromBinary` and `GetELFSlice`.
+  absl::Status LoadBinary(const std::string_view file_name);
+
+  // Returns a pointer inside the loaded binary casted down to an ELF object.
+  absl::StatusOr<llvm::object::ELF64LEObjectFile*> GetELFFromBinary();
```
ondrasej, line 64: Please add a comment who owns the returned pointer (and what is its lifetime).

gematria/datasets/annotating_importer.h
ondrasej: SGTM. Are you planning to do it in this PR or in a follow-up?

gematria/datasets/annotating_importer.h
```diff
@@ -0,0 +1,120 @@
...
+  // Creates a new annotation collector from a given canonicalizer. The
+  // canonicalizer must be for the architecture/microarchitecture of the data
+  // set. Does not take ownership of the canonicalizer.
+  explicit AnnotatingImporter(const Canonicalizer* canonicalizer);
+
+  // Reads an ELF object along with a corresponding `perf.data`-like file and
+  // returns a vector of annotated `BasicBlockProto`s consisting of basic blocks
+  // from the ELF object annotated using samples from the `perf.data`-like file.
+  absl::StatusOr<std::vector<BasicBlockWithThroughputProto>>
+  GetAnnotatedBasicBlockProtos(const std::string_view elf_file_name,
```
ondrasej, line 50: Nit & no action really needed: for arguments passed by value, `const` in a declaration doesn't mean much and the implementation can add or remove it at will (and since the argument is passed by value, whether there is a const or not makes no difference for the caller).

ondrasej (changes requested):
Added a few more comments, but overall it looks good to me. I'll approve once everything is resolved.

ondrasej (approved):

PR #141, opened 2024-07-19
Use `--test_output=errors` when running `bazel test` in CI.

ondrasej:
With this flag, when a test fails, we should see the logs from the failure.


virajbshah (approved):
This should be helpful, thanks!

PR #142, opened 2024-07-21
Cleanup bazel lit tests

boomanaiden154:
This patch cleans up the bazel lit tests to use a standard driver script and a starlark rule that globs for the tests explicitly rather than doing it in an ad-hoc fashion in every individual test definition.


gematria/testing/lit.bzl
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Contains rules for running lit tests."""
+
+load("@bazel_skylib//lib:paths.bzl", "paths")
+
+def lit_test(name, data = []):
```
ondrasej, line 18: Please add a docstring.

Same for the other function.

gematria/testing/lit.bzl
```diff
@@ -0,0 +1,40 @@
...
+        name = name + "_lit_test",
+        srcs = ["@llvm-project//llvm:utils/lit/lit.py"],
+        args = [lit_file_path, "-vv"],
+        main = "@llvm-project//llvm:utils/lit/lit.py",
+        data = [":" + name, "lit.cfg.py", "lit.site.cfg.py"] + data,
+    )
+
+def glob_lit_tests(name, test_file_exts, data = []):
+    test_files = native.glob(
+        ["*." + test_file_exts[0]],
```
ondrasej, line 31: This should either be a list comprehension (if we want to accept multiple extensions) or we should take just one string (if we use only one).

gematria/testing/lit.bzl
```diff
@@ -0,0 +1,40 @@
...
+
+def lit_test(name, data = []):
+    lit_file_path = paths.join(native.package_name(), name)
+
+    native.py_test(
+        name = name + "_lit_test",
+        srcs = ["@llvm-project//llvm:utils/lit/lit.py"],
+        args = [lit_file_path, "-vv"],
+        main = "@llvm-project//llvm:utils/lit/lit.py",
+        data = [":" + name, "lit.cfg.py", "lit.site.cfg.py"] + data,
```
ondrasej, line 26: The dependency on `:{name}` looks suspicious, and almost looks like an implicit circular dependency.

Given how it's called from `glob_lit_tests()`, I think it's OK to keep the current structure, but I'd
1. mark this as private (add a leading underscore),
2. rename the arg `name` to `test_file_name` or something like that to avoid confusion with the usual target attribute naming conventions.

If you'd prefer to keep this public (because it might be called in other ways than through `glob_lit_tests`, then I'd prefer to make `name` the target name, and use `srcs` for the name of the file (to be in line with the usual way of defining targets).

ondrasej (approved):
Cool, this looks much cleaner.

gematria/testing/lit.bzl
boomanaiden154: I've changed it to be private as I don't expect any of the tests within `gematria` to explicitly specify lit targets.

PR #145, opened 2024-07-22
Add name-based lookups to the tflite inference API.

virajbshah:
* Remove the need to know tensor indices at compile time by examining the tensor names used in the tflite models.

### DIFF
--- a/gematria/granite/graph_builder_model_inference.cc
+++ b/gematria/granite/graph_builder_model_inference.cc
@@ -15,12 +15,15 @@
 #include "gematria/granite/graph_builder_model_inference.h"
 
 #include <algorithm>
+#include <array>
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <string_view>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


gematria/granite/graph_builder_model_inference.cc
```diff
@@ -346,7 +424,48 @@ GraphBuilderModelInference::FromTfLiteModel(
...
+  if (llvm::Error error = interpreter.takeError()) return error;
+
+  // Get a mapping between the input tensors used in the model and their
+  // corresponding indices.
+  auto input_tensor_to_idx =
+      std::make_unique<std::vector<int>>(kNumPossibleInputTensors, -1);
+  for (int idx : (*interpreter)->inputs()) {
+    const std::string_view name = (*interpreter)->GetInputName(idx);
+    input_tensor_to_idx->at(static_cast<int>(GetInputTensorFromName(name))) =
+        idx;
```
ondrasej, line 436: We also need to handle the case where the input name is not one of the expected names. The TFLite model is supplied by the user, and it might contain any inputs or outputs, intentionally or by accident.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -412,19 +532,30 @@ GraphBuilderModelInference::FromTfLiteModel(
       *std::move(node_token_list), /* immediate_token = */ *immediate_token,
       /* fp_immediate_token = */ *fp_immediate_token,
       /* address_token = */ *address_token, /* memory_token = */ *memory_token,
+      /* annotation_names */ std::vector<std::string>(),
```
ondrasej, line 535: It looks like this escaped the split.

gematria/granite/graph_builder_model_inference.h
```diff
@@ -87,10 +88,22 @@ class GraphBuilderModelInference {
   // structure of a model based on the BasicBlockGraphBuilder class.
   GraphBuilderModelInference(
       std::unique_ptr<BasicBlockGraphBuilder> graph_builder,
-      const tflite::FlatBufferModel* tflite_model);
+      const tflite::FlatBufferModel* tflite_model,
+      std::unique_ptr<tflite::Interpreter> interpreter,
+      std::unique_ptr<std::vector<int>> input_tensor_to_idx,
```
ondrasej, line 93: Why do we need this to be a unique pointer rather than just a vector? We do not allow `nullptr`, and `std::move()` of a `std::vector` is about as efficient as a move of a `std::unique_ptr<>`.

gematria/granite/graph_builder_model_inference.h
```diff
@@ -87,10 +88,22 @@ class GraphBuilderModelInference {
...
+      bool uses_deltas = true);
```
ondrasej, line 94: We could take it the other way around, and detect `use_deltas` from the model (if there is the delta block mapping, then it uses deltas; otherwise, it doesn't use them).

If we can't auto-detect `use_deltas`, then I'd prefer to have it provided explicitly to avoid any errors where it is skipped by accident.

ondrasej (changes requested):
Cool, thanks for working on this! The name-based matching will make the model construction much simpler and less error-prone 👍

boomanaiden154:
For what it's worth, there are abstractions in upstream LLVM for the TFLite side of things. See `MLModelRunner.h` and `ModelUnderTrainingRunner.h`.

gematria/granite/graph_builder_model_inference.cc
virajbshah: That needs to be there since `BasicBlockGraphBuilder` already supports annotations, so we just pass in an empty vector to match the argument list - but the behaviour should be the same as before annotations.

gematria/granite/graph_builder_model_inference.h
virajbshah: Ah, `FromTfLiteModel` checks for the delta block index tensor and passes `uses_deltas` to the `GraphBuilderModelInference` constructor accordingly. I should remove the default value to avoid confusion.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -346,7 +424,48 @@ GraphBuilderModelInference::FromTfLiteModel(
...
+        idx;
+  }
+
+  // Checks whether the model has all input tensors required to have for all
+  // graph builder models.
+  std::vector<InputTensor> missing_inputs;
+  for (const InputTensor input : kRequiredInputTensors) {
+    if (input_tensor_to_idx->at(static_cast<int>(input)) == -1)
+      missing_inputs.push_back(input);
+  }
```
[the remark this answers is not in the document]
virajbshah: The approach I went for was to make sure the required input tensors were there, and infer the configuration (i.e. `uses_deltas` or not from the presence of the delta block index tensor). So if `kDeltaBlockIndexTensor` is missing. we just assume it is a `seq2num` model and continue. Maybe it would be better to log any such assumptions we make about the configuration, e.g.
```
Found `ModelBase.delta_block_index_tensor`, assuming model uses deltas.
```
Or would that be unnecessarily verbose?

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -346,7 +424,48 @@ GraphBuilderModelInference::FromTfLiteModel(
...
+  const bool uses_deltas = input_tensor_to_idx->at(static_cast<int>(
+                               InputTensor::kDeltaBlockIndexTensor)) != -1;
+  const int num_expected_input_tensors =
+      kNumRequiredInputTensors + int(uses_deltas);
+  if (num_input_tensors != num_expected_input_tensors) {
+    return llvm::createStringError(
+        llvm::errc::invalid_argument,
+        "Unexpected number of input tensors. Expected %d, found %d.",
+        num_expected_input_tensors, num_input_tensors);
+  }
```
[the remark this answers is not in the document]
virajbshah: I've just added that as part of handling the case where the name passed to `GetInputTensorFromName` is not valid, so the input tensor count check is no longer strictly required. But I can leave it in, since it is cheap and seems like a nice exit point after doing all the checks on the model and reporting all of the extra and missing tensors.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -428,25 +424,31 @@ GraphBuilderModelInference::FromTfLiteModel(
...
     const std::string_view name = (*interpreter)->GetInputName(idx);
-    input_tensor_to_idx->at(static_cast<int>(GetInputTensorFromName(name))) =
-        idx;
+    std::optional<InputTensor> input_tensor = GetInputTensorFromName(name);
+
+    // Report any unexpected tensors.
+    if (!input_tensor.has_value()) {
+      llvm::errs() << "Model contains unexpected tensor `" << name
+                   << "` at index " << idx << '\n';
+      continue;
```
ondrasej, line 436: We should probably return an error here (or collect all the unexpected tensor names and then return them all in an error, that's even more useful), because it means that we won't be able to use this model in the inference library (there's no code that would fill in the unknown input tensor).

As a side note: in library code, it's usually better to keep logging light or do it only when requested (and return important information through the status codes, etc.). Otherwise, this quickly makes the logs very hard to read without additional tools for filtering.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -428,25 +424,31 @@ GraphBuilderModelInference::FromTfLiteModel(
...
+    }
+    input_tensor_to_idx.at(static_cast<int>(input_tensor.value())) = idx;
```
ondrasej, line 438: Why not `input_tensor_to_idx[static_cast<int>(input_tensor.value())] = idx` ?

It's more common (and more efficient, because `at()` always does a range check on the index, but that's not something we need to worry about here).

Same for all the other uses of `std::vector<>::at()`.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -346,7 +420,54 @@ GraphBuilderModelInference::FromTfLiteModel(
...
+  }
+
+  // Checks whether the model has all input tensors required to have for all
+  // graph builder models.
+  std::vector<InputTensor> missing_inputs;
+  for (const InputTensor input : kRequiredInputTensors) {
+    if (input_tensor_to_idx.at(static_cast<int>(input)) == -1)
+      missing_inputs.push_back(input);
+  }
```
ondrasej, line 447: I'd add the check for the deltas tensor here, e.g.
```cc
if (uses_deltas && input_tensor_to_idx[InputTensor::kDeltaBlockIndexTensor] == -1) {
  missing_tensor.push_back(InputTensor::kDeltaBlockIndexTensor);
}
```

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -46,20 +50,88 @@ namespace {
...
+enum class InputTensor {
+  kDeltaBlockIndexTensor = 0,
+  kGraphNodesTensor,
+  kGraphEdgesTensor,
+  kGraphGlobalsTensor,
+  kGraphReceiversTensor,
+  kGraphSendersTensor,
+  kGraphNEdgeTensor,
+  kGraphNNodeTensor,
+  kInstructionNodeMaskTensor,
```
ondrasej, line 64: Optional: Since this is an enum class (and thus always requires the `InputTensor::` qualification), I think we can remove the `Tensor` at the end of the value names, and make them a bit shorter.

gematria/granite/graph_builder_model_inference.cc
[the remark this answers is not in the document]
ondrasej: With returning an error for unexpected tensors right above, I think we can remove this part.

Or, maybe even nicer, we can collect missing input tensor names and unexpected input tensor names, and if any of these is non-empty, then return them both in a single error message here.

But if we do the checks above correctly, the checks for the number of tensors would not provide any additional safety.

ondrasej (changes requested):

gematria/granite/graph_builder_model_inference.cc
virajbshah: I'm not sure that sounds right, since `uses_deltas` is defined as
```cpp
const bool uses_deltas = input_tensor_to_idx[static_cast<int>(
                              InputTensor::kDeltaBlockIndexTensor)] != -1;
```
i.e. it's the other way around, we infer what `uses_deltas` should be based on whether we found `kDeltaBlockIndexTensor` or not. I think [this response to a previous comment](https://github.com/google/gematria/pull/145#discussion_r1687962267) is relevant.

ondrasej: Right, we can just skip this part.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -83,21 +83,34 @@ constexpr std::string_view kInstructionNodeMaskTensorName =
...
+// and the third assertion with the last `InputTensor` when a new one is added.
+static_assert(static_cast<int>(kPossibleInputTensors[0].first) == 0,
+              "The start of `kPossibleInputTensors` is not aligned with the "
+              "start of `InputTensor`");
+static_assert(static_cast<int>(InputTensor::kDeltaBlockIndex) == 0,
+              "`InputTensor` does not begin enumerating from 0");
+static_assert(
+    static_cast<int>(InputTensor::kInstructionNodeMask) ==
+        kNumPossibleInputTensors - 1,
+    "`kPossibleInputTensors` does not have an entry for every `InputTensor`");
```
ondrasej, line 112: We could also do something like
```cc
template<typename Container>
constexpr bool IsIncreasing(const Container& entries) {
  const int num_items = std::size(entries);
  for (int i = 1; i < num_items; ++i) {
    if (entries[i - 1].first < entries[i].first) return false;
  }
  return true;
}
static_assert(IsIncreasing(kPossibleInputTensors));
```
to check that the indices are contiguous (there are no duplicates and they are sorted). Together with the other checks, this should cover most of the things that could go wrong.

gematria/granite/graph_builder_model_inference.cc
```diff
@@ -425,49 +438,56 @@ GraphBuilderModelInference::FromTfLiteModel(
...
+        buffer << GetInputTensorName(missing_input) << ", ";
+      }
+      buffer.seekp(-2, std::ios_base::end);
+      buffer << " were expected but not found.";
+      if (!unexpected_input_indices.empty()) {
+        buffer << '\n';
+      }
+    }
+    if (!unexpected_input_indices.empty()) {
+      buffer << "Model is has unexpected extra input tensors. ";
```
ondrasej, line 476: Remove "is".

ondrasej (approved):
Two more comments, otherwise it looks good.

ondrasej (approved):

Issue #147, opened 2024-07-22
convert_bhive_to_llvm_exegesis_inputs tests failing when binary is optimized


boomanaiden154:
I'm observing the following failures:
```
FAIL: //gematria/datasets/convert_bhive_to_llvm_exegesis_input_tests:conversion.test_lit_test (see /root/.cache/bazel/_bazel_root/5b63f27bc35a3d0572c069ebf1768159/execroot/com_google_gematria/bazel-out/k8-opt/testlogs/gematria/datasets/convert_bhive_to_llvm_exegesis_input_tests/conversion.test_lit_test/test.log)
FAIL: //gematria/datasets/convert_bhive_to_llvm_exegesis_input_tests:loop_register.test_lit_test (see /root/.cache/bazel/_bazel_root/5b63f27bc35a3d0572c069ebf1768159/execroot/com_google_gematria/bazel-out/k8-opt/testlogs/gematria/datasets/convert_bhive_to_llvm_exegesis_input_tests/loop_register.test_lit_test/test.log)
FAIL: //gematria/datasets/convert_bhive_to_llvm_exegesis_input_tests:max_bb_count.test_lit_test (see /root/.cache/bazel/_bazel_root/5b63f27bc35a3d0572c069ebf1768159/execroot/com_google_gematria/bazel-out/k8-opt/testlogs/gematria/datasets/convert_bhive_to_llvm_exegesis_input_tests/max_bb_count.test_lit_test/test.log)
```
specifically when I run `blaze test -c opt ...`. These failures notably do not appear when running `blaze test ...`.

PR #149, opened 2024-07-22
Add opt build configuration to CI

boomanaiden154:
This should help catch issues like #147 much more quickly in the future.


ondrasej (approved):
Eventually, we might want to run also ASAN on CI, but this one should be good enough for now.

PR #152, opened 2024-07-23
Add an explicit dependency on `lld` to `Dockerfile`.

ondrasej:
As of 2024-07-23, `bazel build ...` fails unless `lld` is explicitly installed via `apt-get install`.


boomanaiden154 (approved):
LGTM.

Thanks for this. I ran into this issue yesterday when setting up another dev container and meant to send a patch but didn't end up sending one.

PR #153, opened 2024-07-23
Disable pytype module-attr around glorot_normal

boomanaiden154:
Other installations of tensorflow/tf_keras can have complicated `__init__.py`s that pytype isn't able to comprehend correctly. Disable pytype around the one case of this so we can utilize gematria with alternative dependency installations.


gematria/granite/python/token_graph_builder_model.py
```diff
@@ -279,13 +279,15 @@ def _create_readout_network(self) -> tf.Tensor:
   def _create_graph_network_modules(
       self,
   ) -> Sequence[gnn_model_base.GraphNetworkLayer]:
+    # pytype: disable=module-attr
```
ondrasej, line 282: Please add a comment explaining why this is needed.

ondrasej (approved):

PR #155, opened 2024-07-23
Add __builtin_unreachable to end of GetInputTensorName

boomanaiden154:
Currently GetInputTensorName can potentially return from a non-void function without actually returning anything. Add `__builtin_unreachable` to the end of the function to mark that we expect this case to never happen.

### DIFF
--- a/gematria/granite/graph_builder_model_inference.cc
+++ b/gematria/granite/graph_builder_model_inference.cc
@@ -78,7 +78,6 @@ constexpr std::string_view kGraphNEdgeTensorName = "GnnModelBase.num_edges";
 constexpr std::string_view kGraphNNodeTensorName = "GnnModelBase.num_nodes";
 constexpr std::string_view kInstructionNodeMaskTensorName =
     "GraphBuilderModelBase.instruction_node_mask";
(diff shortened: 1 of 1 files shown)


gematria/granite/graph_builder_model_inference.cc
```diff
@@ -128,6 +129,7 @@ constexpr std::string_view GetInputTensorName(const InputTensor tensor) {
       return name;
     }
   }
+  return kUnknownTensorName;
```
ondrasej, line 132: `kPossibleInputTensors` should be exhaustive, so this should never happen. I think we should have an assertion or something similar here.

Either `assert(false && "Unexpected InputTensor value");` or some way to invoke `__builtin_unreachable()`, e.g. `LLVM_BUILTIN_UNREACHABLE`.

ondrasej (approved):

gematria/granite/graph_builder_model_inference.cc
boomanaiden154: I've switched to just using `__builtin_unreachable` directly as I think any of the compilers we will use here should have support for it.

PR #157, opened 2024-07-25
Add beam pipeline for compiling modules

boomanaiden154:
This patch adds an Apache Beam pipeline that can be used for importing LLVM bitcode modules from the ComPile dataset and extracting BBs from them.


boomanaiden154:
This is a stacked PR, based on top of #156.

gematria/datasets/pipelines/compile_modules.py
```diff
@@ -0,0 +1,52 @@
...
+    required=True,
+)
+
+_OUTPUT_TXT_FILE = flags.DEFINE_string(
+    'output_txt_file', None, 'The path to the output txt file', required=True
+)
+
+
+def main(argv) -> None:
+  del argv
```
ondrasej, line 39: Very nit: `del` should always come with a comment. A simple "Unused." is fine.

See https://google.github.io/styleguide/pyguide.html#214-decision for the comment.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+# limitations under the License.
+
+import apache_beam as beam
+from rules_python.python.runfiles import runfiles
+import os
+
+from absl import logging
+
+from collections.abc import Iterable
+from collections.abc import Callable
```
ondrasej, line 22: These two can be on the same line:
```py
from collections.abc import Callable, Iterable
```

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+
+import subprocess
+
+from gematria.datasets.python import extract_bbs_from_obj
```
ondrasej, line 26: The imports should be sorted:
1. standard library imports (`collections.abc`, `os`, `subprocess`),
2. third-party library imports (`absl`, `apache_beam`, `rules_python`),
3. project imports (anything under `gematria`).

See https://google.github.io/styleguide/pyguide.html#313-imports-formatting for the detailed rule.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def _get_llvm_binary_path(tool_name: str) -> str:
+  runfiles_env = runfiles.Create(os.environ)
+  assert runfiles_env is not None
```
ondrasej, line 31: We might want to cache this part (to avoid any expensive operations like extraction or copying files).

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+  return runfiles_env.Rlocation('llvm-project/llvm/' + tool_name)
+
+
+class OptimizeModules(beam.DoFn):
+  """A Beam function that uses LLVM opt to optimize bitcode modules."""
+
+  def __init__(self, optimization_pass_lists: list[str]):
```
ondrasej, line 38: I'd recommend taking a `Sequence[str]` here. If we need a list, then we can do `list(optimization_pass_lists)` below.

ondrasej, line 38: Also, based on how `self.optimization_pass_lists` is used below in `self.process()`, I'd expect it to be `Sequence[Sequence[str]]` rather than just `Sequence[str]`.

Note that `str` is also a `Sequence[str]` (a sequence of characters of the original string), so any type checks around sequences of strings must be treated with extra care.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+    with subprocess.Popen(
+        command_vector,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    ) as opt_process:
+      (output_bc, stderr) = opt_process.communicate(input=input_module)
+      if opt_process.returncode != 0:
+        logging.error(stderr)
+        raise ValueError('Expected opt to return 0')
```
ondrasej, line 55: Same as in the earlier PR: this could be done with a single 
```py
result = subprocess.run(
    stdin=input_module,
    capture_output=False,
)
if result.returncode != 0:
  logging.error(result.stderr)
  raise ValueError(...)
return result.stdout
```

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+    for optimization_pass_list in self.optimization_pass_lists:
+      try:
+        yield self.optimize_module(input_module, optimization_pass_list)
+      except ValueError:
+        continue
+
+
+class LowerModulesAsm(beam.DoFn):
+  """A Beam function that uses LLVM llc to lower bitcode modules to object
+  files.
```
ondrasej, line 68: Nit: This should fit into the 80 columns limit.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+  assert runfiles_env is not None
+  return runfiles_env.Rlocation('llvm-project/llvm/' + tool_name)
+
+
+class OptimizeModules(beam.DoFn):
+  """A Beam function that uses LLVM opt to optimize bitcode modules."""
+
+  def __init__(self, optimization_pass_lists: list[str]):
+    self.optimization_pass_lists = optimization_pass_lists
+    self.opt_path = _get_llvm_binary_path('opt')
```
ondrasej, line 40: Very nit: Let's make these private (by adding a leading underscore).

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class LowerModulesAsm(beam.DoFn):
+  """A Beam function that uses LLVM llc to lower bitcode modules to object
+  files.
+  """
+
+  def __init__(self, optimization_levels: list[str]):
+    self.optimization_levels = optimization_levels
+    self.llc_path = _get_llvm_binary_path('llc')
```
ondrasej, line 73: Same ass above (let's make them "private").

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+        command_vector,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    ) as llc_process:
+      (output_obj, stderr) = llc_process.communicate(input=input_module)
+      if llc_process.returncode != 0:
+        logging.error(stderr)
+        raise ValueError('Expected llc to return 0')
+      return output_obj
```
ondrasej, line 92: Same `subprocess.run()` comment applies here.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+    for bb_hex_value in extract_bbs_from_obj.get_basic_block_hex_values(
+        input_object_file
+    ):
+      yield bb_hex_value
+
+
+class DeduplicateBBs(beam.ptransform.PTransform):
+  """A Beam transform to deduplicate string data."""
+
+  def expand(self, pcoll):
```
ondrasej, line 115: Can we have a more precise type annotation for `pcoll`? If not, please add a docstring explaining what's the expected structure of the pcollection.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -0,0 +1,156 @@
...
+        logging.error(stderr)
+        raise ValueError('Expected opt to return 0')
+      return output_bc
+
+  def process(self, input_module: bytes) -> Iterable[bytes]:
+    for optimization_pass_list in self.optimization_pass_lists:
+      try:
+        yield self.optimize_module(input_module, optimization_pass_list)
+      except ValueError:
+        continue
```
ondrasej, line 63: It could be useful to use state/metrics to track the number of failed modules: https://beam.apache.org/documentation/programming-guide/#combiningstate

Same for the other fn's where we drop modules on error.

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -0,0 +1,122 @@
...
+from absl.testing import absltest
+import os
+
+from gematria.datasets.pipelines import compile_modules_lib
+from gematria.testing.python import ir_utils
+
+from apache_beam.testing.test_pipeline import TestPipeline
+from apache_beam.testing.util import assert_that
+from apache_beam.testing.util import equal_to
+import apache_beam as beam
```
ondrasej, line 24: Same comment about import sorting applies here.

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -0,0 +1,122 @@
...
+
+class CompileModulesTests(absltest.TestCase):
+
+  def test_optimize_modules(self):
+    ir_string = """
+define i32 @a() {
+  %a = add i32 3, 5
+  ret i32 %a
+}
+"""
```
ondrasej, line 35: Nit: Use `textwrap.dedent()` for nicer indentation.

Same below.

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -0,0 +1,122 @@
...
+define i32 @a() {
+  ret i32 8
+}
+"""
+    optimized_modules_text = [
+        ir_utils.get_ir_from_bc(module_bc) for module_bc in optimized_modules
+    ]
+
+    self.assertTrue(ir_string in optimized_modules_text[0])
+    self.assertTrue(optimized_ir_string in optimized_modules_text[1])
```
ondrasej, line 55: `self.assertIn()` for cleaner code and better error messages.

Also, I don't think Beam makes any guarantees about the order of the values in the output. In this case, the order is preserved, but this is more of a coincidence (because the input is small, and there's just one worker running in one thread).

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -0,0 +1,122 @@
...
+
+from absl.testing import absltest
+import os
+
+from gematria.datasets.pipelines import compile_modules_lib
+from gematria.testing.python import ir_utils
+
+from apache_beam.testing.test_pipeline import TestPipeline
+from apache_beam.testing.util import assert_that
+from apache_beam.testing.util import equal_to
```
ondrasej, line 23: The Google style guide does not allow direct imports of classes or functions (with the exception of `collections.abc` and `typing`).

See https://google.github.io/styleguide/pyguide.html#22-imports for more thoughts on this.

ondrasej (changes requested):
Nice!

gematria/datasets/pipelines/compile_modules_lib.py
boomanaiden154: It will be cached by the initialization of the implementation of `beam.DoFn` given it's only called in `__init__` there. I'm not sure caching it anywhere else makes sense as we could have things running across multiple machines where a process like extraction might need to be performed once per machine.

gematria/datasets/pipelines/compile_modules_lib_test.py
boomanaiden154: In general yes, but we're explicitly calling the classes `process` function here, which will just map to the implementation in `compile_modules_lib.py` rather than going through any fancy Beam stuff where it might be reordered.

gematria/datasets/pipelines/compile_modules_lib.py
boomanaiden154: Good point, but it doesn't seem like there's an easy to get the values out after the `DoFn` has been run?

In ordered to read the value of a `CombiningValueStateSpec`, it seems like you need a `DoFn.StateParam` as the parameter to the `process` function of a `DoFn`?

ondrasej (changes requested):
A few more comments, but overall it looks good now.

gematria/datasets/pipelines/compile_modules_lib.py
boomanaiden154: I'll look into at some point. Seems like getting it working in a separate patch would be ideal.

ondrasej (approved):

gematria/datasets/pipelines/compile_modules_lib.py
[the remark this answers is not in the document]
boomanaiden154: Done. I would've thought `pyink` handled this natively. Apparently not. Will keep an eye out for it in the future.

PR #158, opened 2024-07-25
Pin LLVM version used by CMake CI.

virajbshah:
* The LLVM version used by the CMake CI is now pinned to that specified by `WORKSPACE`.
 * It must be manually bumped alongside the one in `WORKSPACE`.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,6 +14,9 @@
 
 name: Gematria CI
 
+env:
+  LLVM_COMMIT: 74a5e7784b32aba5670ff427b158d1e6e38012f1
+
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -14,6 +14,9 @@
 
 name: Gematria CI
 
+env:
+  LLVM_COMMIT: 74a5e7784b32aba5670ff427b158d1e6e38012f1
```
ondrasej, line 18: Ideally, there should be just one version info that can be read both from here and from `WORKSPACE`. But since it might take some time to figure out how to do that correctly, let's:
- add a comment here and in the `WORKSPACE` file that points to the other spot and says that the versions need to be kept in sync.
- add a TODO here (for me or for you) to figure out how to have just one LLVM version stored somewhere in the repository. I guess in the worst case, we could `grep` it out of `WORKSPACE` in here.

ondrasej (approved):

PR #159, opened 2024-07-25
Fix llvm-cm tests against recent LLVM

boomanaiden154:
Bumping to a recent LLVM version causes the LLVM tests to fail. This patch updates the tests and makes them slightly more resilient against small changes.

### DIFF
--- a/llvm_cm/test/X86/multi_func.s
+++ b/llvm_cm/test/X86/multi_func.s
@@ -5,31 +5,31 @@
 
 
 # CHECK:      <reverse>:
-# CHECK-NEXT: Calculated Frequency: 8.3{{[0-9]+}}e+03
+# CHECK-NEXT: Calculated Frequency: 8.{{[0-9]+}}e+03
 # CHECK-NEXT: <tallestBillboard>:
-# CHECK-NEXT: Calculated Frequency: 2.9{{[0-9]+}}e+05
+# CHECK-NEXT: Calculated Frequency: 2.{{[0-9]+}}e+05
 # CHECK-NEXT: <isMatch>:
-# CHECK-NEXT: Calculated Frequency: 8.2{{[0-9]+}}e+02
+# CHECK-NEXT: Calculated Frequency: 8.{{[0-9]+}}e+02
 # CHECK-NEXT: <bubbleSort>:
-# CHECK-NEXT: Calculated Frequency: 6.6{{[0-9]+}}e+04
+# CHECK-NEXT: Calculated Frequency: 6.{{[0-9]+}}e+04
 # CHECK-NEXT: <isPrime>:
-# CHECK-NEXT: Calculated Frequency: 5.7{{[0-9]+}}e+02
+# CHECK-NEXT: Calculated Frequency: 5.{{[0-9]+}}e+02
 # CHECK-NEXT: <main>:
-# CHECK-NEXT: Calculated Frequency: 8.8{{[0-9]+}}e+03
+# CHECK-NEXT: Calculated Frequency: 8.{{[0-9]+}}e+03
 
 
 # CHECK-COUNT: <reverse>:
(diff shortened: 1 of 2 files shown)


virajbshah:
I think this one needs some more investigating. It looks like one of the expected outputs is off by a good bit, and we would want to pin the LLVM version to one where things work as intended (like #158 does for the CI with the version in `WORKSPACE`).

For what it's worth, some `git bisect run`-ing tells us that the LLVM commit that broke the tests was `ea222be0d9266d9d5c100496f8c9606f213454ee`.

boomanaiden154:
> I think this one needs some more investigating. It looks like one of the expected outputs is off by a good bit, and we would want to pin the LLVM version to one where things work as intended (like https://github.com/google/gematria/pull/158 does for the CI with the version in WORKSPACE).

We can't just pin the LLVM version to one where things work as expected. That prevents us from importing the code into google3 (at least without test failures) as the LLVM version there essentially tracks ToT, so we need to try and follow close to ToT externally.

> I think this one needs some more investigating. It looks like one of the expected outputs is off by a good bit, and we would want to pin the LLVM version to one where things work as intended (like https://github.com/google/gematria/pull/158 does for the CI with the version in WORKSPACE).

For what it's worth, some git bisect run-ing tells us that the LLVM commit that broke the tests was ea222be0d9266d9d5c100496f8c9606f213454ee.

Thanks for the git bisect. Based on that commit, the changes are pretty likely to be noise. Given that `.p2align` expands differently, some difference in the counts is expected.

The tests need to be rewritten at some point to not just contain a dump of assembly so that they vary less from changes like this. I think IR with profile metadata would be the optimal choice but rewriting them has been low priority.

ondrasej:
> We can't just pin the LLVM version to one where things work as expected. That prevents us from importing the code into google3 (at least without test failures) as the LLVM version there essentially tracks ToT, so we need to try and follow close to ToT externally.

I'm rather for pinning the version:
- When we pin, we might be lagging behind LLVM head a bit, but at least we can always have CI passing, and a project that builds.
- Without it, our build and CI might get broken by random (well meaning) changes in LLVM and we'd only find out when the project doesn't work. Which means that it can block others from working, add noise to CI, ...
- Even if we pin, it doesn't mean that we need to stay on one version for too long, we can continuously keep moving forward similarly to what google3 does. But we'd be doing it in a controlled manner (1) when someone has time to work on broken tests, and (2) while keeping our CI status green.

ondrasej:
> For what it's worth, some git bisect run-ing tells us that the LLVM commit that broke the tests was ea222be0d9266d9d5c100496f8c9606f213454ee.
> 
> Thanks for the git bisect. Based on that commit, the changes are pretty likely to be noise. Given that `.p2align` expands differently, some difference in the counts is expected.
> 
> The tests need to be rewritten at some point to not just contain a dump of assembly so that they vary less from changes like this. I think IR with profile metadata would be the optimal choice but rewriting them has been low priority.

When merging, please add all this info to the commit message for the record.

ondrasej (approved):
Approve to unblock the fix, but we should continue the discussion on LLVM version pinning.

boomanaiden154:
> I'm rather for pinning the version:

My last response I guess misses this a bit. I'm definitely for pinning here, just trying to point out that we do need to roll past any breaking changes to keep the google3 import working.

> When we pin, we might be lagging behind LLVM head a bit, but at least we can always have CI passing, and a project that builds. Without it, our build and CI might get broken by random (well meaning) changes in LLVM and we'd only find out when the project doesn't work. Which means that it can block others from working, add noise to CI, ...

Yep.

> Even if we pin, it doesn't mean that we need to stay on one version for too long, we can continuously keep moving forward similarly to what google3 does. But we'd be doing it in a controlled manner (1) when someone has time to work on broken tests, and (2) while keeping our CI status green.

I think we're all on the same page here. I just need to state my thoughts a little bit more clearly next time.

PR #160, opened 2024-07-26
Bump LLVM version

boomanaiden154:
This patch bumps the LLVM version to
67a55e01e3f13d6ea5be917765a4171cd68cb5ac, primarily to bump past the change that broke the llvm-cm tests that were subsequently fixed against the new version.


ondrasej (approved):

ondrasej:
Ideally, this should have gone together with #159, but 🤷  :)

boomanaiden154:
> Ideally, this should have gone together with https://github.com/google/gematria/pull/159, but 🤷 :)

Yep. Didn't see that until I had already hit merge. I'll make sure the CI passes in the PR and then merge.

PR #161, opened 2024-07-28
Refactor process_and_filter_bbs into a library

boomanaiden154:
This will enable the addition of unittests to verify the functionality of the script along with the addition of python bindings so that this tooling can be easily used from within a beam pipeline.

### DIFF
--- a/gematria/datasets/process_and_filter_bbs_lib.cc
+++ b/gematria/datasets/process_and_filter_bbs_lib.cc
@@ -24,12 +24,11 @@ using namespace llvm;
 
 namespace gematria {
 
-BBProcessorFilter::BBProcessorFilter() {
-  LLVMSupport = LlvmArchitectureSupport::X86_64();
-  InstructionPrinter = LLVMSupport->CreateMCInstPrinter(0);
-}
+BBProcessorFilter::BBProcessorFilter()
+    : LLVMSupport(LlvmArchitectureSupport::X86_64()),
+      InstructionPrinter(LLVMSupport->CreateMCInstPrinter(0)) {}
 
-Expected<std::string> BBProcessorFilter::processBasicBlock(
+Expected<std::string> BBProcessorFilter::removeRiskyInstructions(
     const StringRef BasicBlock, const StringRef Filename,
     bool FilterMemoryAccessingBlocks) {
   // TODO(boomanaiden154): Update this to use llvm::Expected once
--- a/gematria/datasets/process_and_filter_bbs_lib.cc
+++ b/gematria/datasets/process_and_filter_bbs_lib.cc
@@ -14,6 +14,15 @@
 
 #include "gematria/datasets/process_and_filter_bbs_lib.h"
 
+#include <string>
+#include <system_error>
+#include <vector>
+
+#include "/gematria/external/llvm-project/llvm/include/llvm/ADT/StringRef.h"
+#include "/gematria/external/llvm-project/llvm/include/llvm/ADT/Twine.h"
+#include "/gematria/external/llvm-project/llvm/include/llvm/MC/MCInstrDesc.h"
+#include "/gematria/external/llvm-project/llvm/include/llvm/Support/Error.h"
+#include "/gematria/external/llvm-project/llvm/lib/Target/X86/MCTargetDesc/X86MCTargetDesc.h"
 #include "X86InstrInfo.h"
 #include "gematria/llvm/disassembler.h"
 #include "gematria/llvm/llvm_architecture_support.h"
--- a/gematria/datasets/process_and_filter_bbs_lib.cc
+++ b/gematria/datasets/process_and_filter_bbs_lib.cc
@@ -18,16 +18,16 @@
 #include <system_error>
 #include <vector>
 
-#include "/gematria/external/llvm-project/llvm/include/llvm/ADT/StringRef.h"
-#include "/gematria/external/llvm-project/llvm/include/llvm/ADT/Twine.h"
-#include "/gematria/external/llvm-project/llvm/include/llvm/MC/MCInstrDesc.h"
-#include "/gematria/external/llvm-project/llvm/include/llvm/Support/Error.h"
-#include "/gematria/external/llvm-project/llvm/lib/Target/X86/MCTargetDesc/X86MCTargetDesc.h"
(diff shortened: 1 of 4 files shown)


ondrasej (changes requested):
We should have at least a small (unit) test for the new library.

ondrasej:
> We should have at least a small (unit) test for the new library.

Actually, unit tests are in #162.

boomanaiden154:
> Actually, unit tests are in https://github.com/google/gematria/pull/162.

Yes. I put them there as this change is NFC and figured it would be easier to review new unit tests separately.

gematria/datasets/process_and_filter_bbs_lib.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <string>
+
+#include "gematria/llvm/llvm_architecture_support.h"
+
+namespace gematria {
+
+class BBProcessorFilter {
+ public:
+  explicit BBProcessorFilter();
```
ondrasej, line 23: The `explicit` has no effect when there are no arguments.

gematria/datasets/process_and_filter_bbs_lib.h
```diff
@@ -0,0 +1,34 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+
+#include "gematria/llvm/llvm_architecture_support.h"
```
ondrasej, line 17: We should also have includes for other types used in the header, in particular `std::unique_ptr`, `llvm::StringRef`, `llvm::Expected`, and `llvm::MCInstPrinter`.

In general, header files should be self-container and carry all the includes or forward declarations needed to make the header parse and compile all by itself.

gematria/datasets/process_and_filter_bbs_lib.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "llvm/ADT/StringExtras.h"
+
+using namespace llvm;
+
+namespace gematria {
+
+BBProcessorFilter::BBProcessorFilter() {
+  LLVMSupport = LlvmArchitectureSupport::X86_64();
+  InstructionPrinter = LLVMSupport->CreateMCInstPrinter(0);
+}
```
ondrasej, line 30: Why not use the initializer syntax?
```cc
BBProcessorFilter::BBProcessorFilter()
    : LLVMSuppert(LlvmArchitectureSupport::X86_64()),
      InstructionPrinter(LLVMSupport->CreateMCInstPrinter(0)) {}
```

gematria/datasets/process_and_filter_bbs_lib.cc
```diff
@@ -0,0 +1,76 @@
...
+
+  std::string OutputBlock;
+
+  for (const gematria::DisassembledInstruction &Instruction :
+       *DisassembledInstructionsOrErr) {
+    MCInstrDesc InstDesc =
+        LLVMSupport->mc_instr_info().get(Instruction.mc_inst.getOpcode());
+    // TODO(boomanaiden154): This filtering is a bit simplistic currently. We
+    // should probably be using MCInsrtDesc::hasUnmodeledSideEffects, but this
+    // needs to be evaluated at scale.
```
ondrasej, line 63: In general, +1 for `hasUnmodeledSideEffects`, but at some point in the past IIRC it covered also instructions like `rep movs` that we might want to model.

gematria/datasets/process_and_filter_bbs_lib.cc
```diff
@@ -0,0 +1,76 @@
...
+using namespace llvm;
+
+namespace gematria {
+
+BBProcessorFilter::BBProcessorFilter() {
+  LLVMSupport = LlvmArchitectureSupport::X86_64();
+  InstructionPrinter = LLVMSupport->CreateMCInstPrinter(0);
+}
+
+Expected<std::string> BBProcessorFilter::processBasicBlock(
```
ondrasej, line 32: We should rename this to `removeRiskyInstructions()` so that the name describes more precisely what the function does.

Also, in the future we might also want to do other things here, e.g. break things down into multiple basic blocks rather than just remove the instructions and we might need multiple filtering functions with different, more descriptive, names.

gematria/datasets/process_and_filter_bbs_lib.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace gematria {
+
+class BBProcessorFilter {
```
ondrasej, line 21: Please add a explaining what the class is for, and also a comment for non-trivial members (right now, probably just `processBasicBlock`).

ondrasej (changes requested):

gematria/datasets/process_and_filter_bbs_lib.h
boomanaiden154: The header will compile by itself, just getting everything from `gematria/llvm/llvm_architecture_support.h"

We should probably add checks for this using something like `clang-tidy`? I think if we have a `clang-tidy` config, `clangd` will pick it up for editor integration.

gematria/datasets/process_and_filter_bbs_lib.cc
boomanaiden154: Right. Punting it off for now given discussion in previous patches.

This was just mechanically copied over from the executable.

gematria/datasets/process_and_filter_bbs_lib.h
ondrasej: The compilation is checked implicitly by this header being the first `#include` in the corresponding `.cc` file (so, that is already done).

But if there's an easy way to enable an include-what-you-use clang-tidy check in editors and in CI, that would be great.

ondrasej (approved):

gematria/datasets/process_and_filter_bbs_lib.h
boomanaiden154: Yep. I think trying to match the internal checks would be good. #164 on the IDE side, I'll look into `clang-tidy` in CI at some point given it needs bazel integration.

PR #163, opened 2024-07-28
Add python bindings for BBProcessorFilter

boomanaiden154:
This patch adds python bindings for BBProcessorFilter so that the tooling can be used within a Beam pipeline to process the basic blocks prior to annotation.


boomanaiden154:
This is a stacked PR on #161.

gematria/datasets/python/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,66 @@
+// Copyright 2023 Google Inc.
```
ondrasej, line 1: 2024

gematria/datasets/python/process_and_filter_bbs.cc
```diff
@@ -0,0 +1,66 @@
...
+      .def(
+          "remove_risky_instructions",
+          [](BBProcessorFilter& Self, std::string BasicBlock,
+             std::string Filename, bool FilterMemoryAccessingBlocks) {
+            return LlvmExpectedToStatusOr(Self.removeRiskyInstructions(
+                BasicBlock, Filename, FilterMemoryAccessingBlocks));
+          },
+          py::arg("basic_block"), py::arg("file_name"),
+          py::arg("filter_memory_accessing_blocks"),
+          R"(Processes a raw basic block extracted from a ELF file.
```
ondrasej, line 44: Very nit: a**n** ELF file

gematria/datasets/python/process_and_filter_bbs_test.py
```diff
@@ -0,0 +1,34 @@
...
+  def setUp(self):
+    super().setUp()
+
+    self.bb_processor_filter = process_and_filter_bbs.BBProcessorFilter()
+
+  def test_bb_processor(self):
+    processed_bb = self.bb_processor_filter.remove_risky_instructions(
+        "B801000000C3", "test", False
+    )
+    self.assertEqual(processed_bb, "B801000000")
```
ondrasej, line 30: We should also have at least one test that raises the exception.

ondrasej (changes requested):

ondrasej (approved):

PR #164, opened 2024-07-31
Add MissingIncludes to clangd config

boomanaiden154:
This patch adds the MissingIncludes option to the clangd config so that IWYU violations are warned about in the IDE, assuming it uses clangd as the LSP.


ondrasej (approved):
Nice! Let's see how this helps us keep the includes in check.

PR #166, opened 2024-08-01
Move loop register into BlockAnnotations

boomanaiden154:
convert_bhive_to_llvm_exegesis_input before this patch stored the loop_register separately. Having the annotators add the loop register to the annotations gives more flexibility down the line for exactly specifying how the block was executed in the annotator/should be benchmarked. Additionally, a register chosen for use as a loop counter is fundamentally an annotation, so placing it there makes more sense from the abstraction perspective.


boomanaiden154:
This is a stacked PR on top of #165.

ondrasej (approved):

PR #168, opened 2024-08-01
Add unittests for BHiveToExegesis

boomanaiden154:
This patch adds unittests for the BHiveToExegesis to ensure that the plumbing works as expected.

The functionality is not tested particularly exhaustively, but the infrastructure is in place to add new tests to prevent regressions in the future, and most of the critical functionality, like detecting memory accesses, is already present elsewhere.


boomanaiden154:
This is a stacked PR:
1. #167
2. #166
3. #165

ondrasej (approved):

PR #171, opened 2024-08-07
Add python bindings for bhive_to_exegesis

boomanaiden154:
This patch adds python bindings for the bhive_to_exegesis library. This enables us to get annotated basic blocks from raw hex values. Future patches will refactor exegesis_benchmark into a library and add additional functions that will take in the AnnotatedBlock class so that we can then get throughput information.

### DIFF
--- a/gematria/datasets/python/bhive_to_exegesis.cc
+++ b/gematria/datasets/python/bhive_to_exegesis.cc
@@ -1,0 +1,70 @@
+// Copyright 2024 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


gematria/datasets/python/bhive_to_exegesis.cc
```diff
@@ -0,0 +1,92 @@
...
+#include "pybind11/stl.h"  // IWYU pragma: keep
+#include "pybind11_abseil/import_status_module.h"
+#include "pybind11_abseil/status_casters.h"  // IWYU pragma: keep
+
+namespace gematria {
+
+namespace py = ::pybind11;
+
+PYBIND11_MODULE(bhive_to_exegesis, m) {
+  m.doc() = "Code for annotating baisc blocks.";
```
orodley, line 35: typo

orodley (approved):

PR #174, opened 2024-08-11
Add unittests for exegesis_benchmark_lib

boomanaiden154:
This patch adds unit tests for the exegesis benchmark library.

### DIFF
--- a/gematria/datasets/exegesis_benchmark_lib_test.cc
+++ b/gematria/datasets/exegesis_benchmark_lib_test.cc
@@ -17,6 +17,7 @@
 #include <memory>
 #include <string>
 
+#include "gematria/testing/llvm.h"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "llvm/ADT/StringRef.h"
@@ -34,6 +35,8 @@ namespace {
 
 using namespace llvm;
 using namespace llvm::exegesis;
+
+using ::testing::UnorderedElementsAre;
 
 class ExegesisBenchmarkTest : public testing::Test {
  protected:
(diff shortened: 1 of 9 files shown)


gematria/datasets/exegesis_benchmark_lib_test.cc
```diff
@@ -0,0 +1,386 @@
...
+    ]
+  }
+  )json";
+
+  Expected<json::Value> BlockValue = json::parse(JSONBlock);
+  ASSERT_TRUE(static_cast<bool>(BlockValue));
+  Expected<BenchmarkCode> BenchCode =
+      Benchmark->parseJSONBlock(*BlockValue->getAsObject(), 1);
+
+  ASSERT_THAT(BenchCode->Key.Instructions, testing::SizeIs(1));
```
orodley, line 127: Consider using some of the container matchers rather than asserting on the size and then pulling out elements.

gematria/datasets/exegesis_benchmark_lib_test.cc
```diff
@@ -0,0 +1,386 @@
...
+
+  ASSERT_THAT(BenchCode->Key.RegisterInitialValues, testing::SizeIs(2));
+  EXPECT_EQ(BenchCode->Key.RegisterInitialValues[0].Register, 54);
+  EXPECT_EQ(BenchCode->Key.RegisterInitialValues[0].Value, 86016);
+  EXPECT_EQ(BenchCode->Key.RegisterInitialValues[1].Register, 60);
+  EXPECT_EQ(BenchCode->Key.RegisterInitialValues[1].Value, 86016);
+}
+
+TEST_F(ExegesisBenchmarkTest, TestParseJSONNoHex) {
+  Expected<std::string> ErrorMessage = getErrorMessage(R"json(
```
orodley, line 144: Nit: making this a multiline string seems unnecessary, I think `getErrorMessage("{}")` would be more readable.

orodley (approved):

PR #175, opened 2024-08-11
Add support for creating a benchmark configuration from annotations

boomanaiden154:
This patch adds functionality in exegesis_benchmark_lib to directly create a benchmark configuration (BenchmarkCode) from annotations (BlockAnnotations). This is intended to be used primarily from the python bindings, where we get annotations directly from the annotator and then need to run benchmarking directly on that rather than serializing to JSON.

### DIFF
--- a/gematria/datasets/exegesis_benchmark_lib.cc
+++ b/gematria/datasets/exegesis_benchmark_lib.cc
@@ -18,9 +18,11 @@
 #include <cstdint>
 #include <memory>
 #include <optional>
+#include <string_view>
 #include <utility>
 #include <vector>
 
+#include "gematria/datasets/find_accessed_addrs.h"
 #include "gematria/llvm/disassembler.h"
 #include "gematria/utils/string.h"
 #include "llvm/ADT/APInt.h"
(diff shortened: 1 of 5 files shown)


gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -261,6 +263,69 @@ Expected<BenchmarkCode> ExegesisBenchmark::parseJSONBlock(
   return BenchCode;
 }
 
+Expected<BenchmarkCode> ExegesisBenchmark::processAnnotatedBlock(
+    std::string_view BlockHex, const BlockAnnotations &Annotations) {
+  std::optional<std::vector<uint8_t>> BytesOr =
```
orodley, line 268: We generally avoid using these kind of "Or" / "OrErr" suffixes. See go/totw/181

gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -261,6 +263,69 @@ Expected<BenchmarkCode> ExegesisBenchmark::parseJSONBlock(
...
+       *InstructionsOrErr)
+    Instructions.push_back(Instruction.mc_inst);
+
+  BenchmarkConfiguration.Key.Instructions = std::move(Instructions);
+
+  BenchmarkConfiguration.Key.RegisterInitialValues.reserve(
+      Annotations.initial_regs.size());
+
+  for (const RegisterAndValue &RegisterValue : Annotations.initial_regs) {
+    struct RegisterValue ValueToAdd = {
```
orodley, line 299: I believe the `struct` tag is unnecessary here

gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -261,6 +263,69 @@ Expected<BenchmarkCode> ExegesisBenchmark::parseJSONBlock(
...
+
+  MemoryValue MemVal = {.Value = APInt(64, Annotations.block_contents),
+                        .SizeBytes = Annotations.block_size,
+                        .Index = 0};
+  BenchmarkConfiguration.Key.MemoryValues["MEM"] = std::move(MemVal);
+
+  for (const uintptr_t AccessedBlock : Annotations.accessed_blocks) {
+    MemoryMapping MemMap = {.Address = AccessedBlock, .MemoryValueName = "MEM"};
+
+    BenchmarkConfiguration.Key.MemoryMappings.push_back(std::move(MemMap));
```
orodley, line 315: If we're reserving space in vectors, may as well do so for `MemoryMappings` too.

gematria/datasets/exegesis_benchmark_lib_test.cc
```diff
@@ -382,5 +384,66 @@ TEST_F(ExegesisBenchmarkTest, TestBenchmarkAdd) {
...
+      .accessed_blocks = {0xaa, 0xbb},
+      .initial_regs = {{.register_index = 0, .register_value = 5},
+                       {.register_index = 1, .register_value = 17}},
+      .loop_register = std::nullopt};
+
+  Expected<BenchmarkCode> BenchmarkConfiguration =
+      Benchmark->processAnnotatedBlock("3b31", Annotations);
+  ASSERT_TRUE(static_cast<bool>(BenchmarkConfiguration));
+
+  ASSERT_THAT(BenchmarkConfiguration->Key.Instructions, testing::SizeIs(1));
```
orodley, line 401: Consider using container matchers

orodley (approved):

gematria/datasets/exegesis_benchmark_lib.cc
boomanaiden154: It's needed here as `RegisterValue` is just a normal struct rather than being `typedef`ed or being a `enum class`. Might be something that should be fixed upstream though...

orodley: This is only the case in C, not C++. C++ doesn't have tag namespaces. The struct tag is accepted in this context, but only for C compatibility.

boomanaiden154: Ah.

I was getting a compile error when trying to get that to work, but looks like I should've read the whole thing:
```
gematria/datasets/exegesis_benchmark_lib.cc:299:5: error: must use 'struct' tag to refer to type 'RegisterValue' in this scope
    RegisterValue ValueToAdd = {
    ^
    struct 
gematria/datasets/exegesis_benchmark_lib.cc:298:32: note: struct 'RegisterValue' is hidden by a non-type declaration of 'RegisterValue' here
  for (const RegisterAndValue &RegisterValue : Annotations.initial_regs) {
```

It was only necessary because of the variable already named `RegisterValue`.

orodley: Huh, TIL something new about C++ shadowing. I thought `struct` was simply ignored in that context, I didn't know it could be used to disambiguate between types and locals that shadow the same name. In C types (whether in the tag namespaces or the typedef namespace) are in an entirely different namespace from variables, so this works very differently.

Some experiments: https://godbolt.org/z/qK65xeoq6
Pretty weird stuff :)

boomanaiden154: Interesting. Today I learned about C++ shadowing. Thanks for digging into that and showing your experiments!

PR #176, opened 2024-08-12
Add python bindings for exegesis_benchmark

boomanaiden154:
This patch adds python bindings to exegesis_benchmark. This allows for benchmarking to be performed from Python, primarily in Beam pipelines.

### DIFF
--- a/gematria/datasets/exegesis_benchmark_lib.cc
+++ b/gematria/datasets/exegesis_benchmark_lib.cc
@@ -265,29 +265,29 @@ Expected<BenchmarkCode> ExegesisBenchmark::parseJSONBlock(
 
 Expected<BenchmarkCode> ExegesisBenchmark::processAnnotatedBlock(
     std::string_view BlockHex, const BlockAnnotations &Annotations) {
-  std::optional<std::vector<uint8_t>> BytesOr =
+  std::optional<std::vector<uint8_t>> Bytes =
       gematria::ParseHexString(BlockHex);
 
-  if (!BytesOr.has_value())
+  if (!Bytes.has_value())
     return llvm::make_error<StringError>(
         errc::invalid_argument,
         "Malformed basic block: invalid hex value " + BlockHex);
 
-  Expected<std::vector<gematria::DisassembledInstruction>> InstructionsOrErr =
-      gematria::DisassembleAllInstructions(
+  Expected<std::vector<gematria::DisassembledInstruction>>
+      DisassembledInstructions = gematria::DisassembleAllInstructions(
           *LLVMMCDisassembler, ExegesisState.getInstrInfo(),
           ExegesisState.getRegInfo(), ExegesisState.getSubtargetInfo(),
-          *LLVMMCInstPrinter, 0, *BytesOr);
+          *LLVMMCInstPrinter, 0, *Bytes);
 
-  if (!InstructionsOrErr) return InstructionsOrErr.takeError();
+  if (!DisassembledInstructions) return DisassembledInstructions.takeError();
 
   BenchmarkCode BenchmarkConfiguration;
 
   std::vector<MCInst> Instructions;
-  Instructions.reserve(InstructionsOrErr->size());
+  Instructions.reserve(DisassembledInstructions->size());
 
   for (const gematria::DisassembledInstruction &Instruction :
-       *InstructionsOrErr)
+       *DisassembledInstructions)
     Instructions.push_back(Instruction.mc_inst);
 
   BenchmarkConfiguration.Key.Instructions = std::move(Instructions);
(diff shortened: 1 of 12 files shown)


gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -0,0 +1,70 @@
...
+             // LLVM Setup
+             LLVMInitializeX86TargetInfo();
+             LLVMInitializeX86TargetMC();
+             LLVMInitializeX86Target();
+             LLVMInitializeX86AsmPrinter();
+             LLVMInitializeX86AsmParser();
+             LLVMInitializeX86Disassembler();
+
+             // Exegesis Setup
+             InitializeX86ExegesisTarget();
```
ondrasej, line 55: I'd prefer to put some more protection around this: create a function that calls the init only once, similar to [`InitializeLlvmOnce()`](https://github.com/google/gematria/blob/21db9938a55255a432db6c7f2a2285448de035a6/gematria/llvm/llvm_architecture_support.cc#L37). This guarantees that the initialization is done only once, and there are no race conditions.

I'm not sure these functions can be called multiple times (likely they can), and I'm very sure they are not thread safe (been there, done that, spent a lot of time debugging).

gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -0,0 +1,70 @@
...
+
+             if (pfm::pfmInitialize())
+               return absl::InternalError("Failed to initialize PFM");
```
ondrasej, line 58: Same here. I guess we can use the same function for the initialization of both.

gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -0,0 +1,70 @@
...
+
+             return LlvmExpectedToStatusOr(ExegesisBenchmark::create());
+           })
+      .def("process_annotated_block",
+           [](ExegesisBenchmark& Self, std::string_view BlockHex,
+              const BlockAnnotations& Annotations) {
+             return LlvmExpectedToStatusOr(
+                 Self.processAnnotatedBlock(BlockHex, Annotations));
```
ondrasej, line 66: Is there a binding for `llvm::exegesis::BenchmarkCode`? If not, do we need one?

gematria/datasets/python/exegesis_benchmark.cc
ondrasej, line 1: We should also have a test for the python code. Nothing complicated, just to check that it loads, initializes, and runs (and returns some reasonable results).

With pybind, a lot of the erros may happen only at run time.

gematria/datasets/exegesis_benchmark_lib.h
```diff
@@ -0,0 +1,61 @@
...
+#include "llvm/tools/llvm-exegesis/lib/LlvmState.h"
+
+namespace gematria {
+
+// ExegesisBenchmark is used for generating benchmark configurations for
+// benchmarks and executing those benchmarks to get real-world data that
+// can then be used for training Gematria models.
+class ExegesisBenchmark {
+ private:
+  explicit ExegesisBenchmark(llvm::exegesis::LLVMState &&State);
```
ondrasej, line 37: Do we need to take the state by rvalue reference? This might complicate memory management in the Python binding (I'm not sure how well a move translates, and it might create dangling pointers or accessible moved objects).

I think we should use raw pointers, and something like https://pybind11.readthedocs.io/en/stable/advanced/functions.html#keep-alive to make sure that the Python object doesn't get garbage collected.

gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -0,0 +1,371 @@
...
+#include "llvm/tools/llvm-exegesis/lib/SnippetRepetitor.h"
+#include "llvm/tools/llvm-exegesis/lib/Target.h"
+
+using namespace llvm;
+using namespace llvm::exegesis;
+
+namespace gematria {
+
+ExegesisBenchmark::ExegesisBenchmark(LLVMState &&State)
+    : ExegesisState(std::move(State)) {
```
ondrasej, line 55: **tl;dr:** The `std::move()` is not needed.

**Longer:** The only thing that `std::move()` does is that it converts a reference to an rvalue reference. When `State` is already an rvalue-reference, this is a no-op.

ondrasej (changes requested):
First batch of comments. I still need to take a look at the `_lib.cc` and `_lib_test.cc`.

gematria/datasets/exegesis_benchmark_lib.h
boomanaiden154: This constructor is private and only called from within the C++, particularly the `create` function, so I don't think we should have any problems with memory management here.

gematria/datasets/python/exegesis_benchmark.cc
boomanaiden154: Added a simple test.

gematria/datasets/python/exegesis_benchmark.cc
boomanaiden154: Added a binding. For now, I didn't add any methods as the python side shouldn't need to access anything inside.

gematria/datasets/exegesis_benchmark_lib.cc
boomanaiden154: Will address this separately given the patch for that already landed.

boomanaiden154: Actually, when trying to fix that, `clangd` is giving me an error about a deleted copy constructor. 

Not entirely sure why given the `std::move` is necessary here given we're basically `static_cast<>`ing a r-value reference to a r-value reference (like you mentioned).

ondrasej (approved):

gematria/datasets/exegesis_benchmark_lib.cc
ondrasej: Ah, right.

From https://en.cppreference.com/w/cpp/language/value_category#lvalue:
> Even if the variable's type is rvalue reference, the expression consisting of its name is an lvalue expression

boomanaiden154: Today I learned. C++ being complicated again. Thanks for pulling that up!

PR #178, opened 2024-08-21
Use np.ndarray.tolist() for numpy arrays.

ondrasej:
Needed to keep in sync with internal cleanups.


boomanaiden154 (approved):
LGTM.

Do you have more context on why exactly this is needed?

ondrasej:
> LGTM.
> 
> Do you have more context on why exactly this is needed?

It's part of numpy code cleanup to be ready for numpy 2.x.

PR #180, opened 2024-08-27
Replace `tf.contrib.training` usage with `tf_slim.evaluation`.

virajbshah:
* `tf.contrib` was dropped with TensorFlow 2 and is not even available under `tf.compat.v1`.
 * The functionality from `tf.contrib.training` that we used is now available in `tf_slim.evaluation`.


ondrasej (approved):
Nice! We should check out what other bits from `tf-lite` we could reuse in our training loop.

PR #185, opened 2024-09-03
Add docstring for get_bbs

boomanaiden154:
This patch adds a docstring for get_bbs. This function is long enough that having a docstring is probably useful. The internal linting tools also complain about a lack of a docstring.

### DIFF
--- a/gematria/datasets/pipelines/compile_modules_lib.py
+++ b/gematria/datasets/pipelines/compile_modules_lib.py
@@ -112,11 +112,11 @@ def get_bbs(
   """Creates a pipeline to process BBs from IR modules.
 
   This function returns a function that builds a beam pipeline to automatically
-  load ir files from a ComPile style parquet file, process them into assembly
+  load IR files from a ComPile style Parquet file, process them into assembly
   basic blocks, deduplicate them, and then write them to a text file.
 
   Args:
-    input_file_pattern: A grep-like pattern to use to search for the parquet
+    input_file_pattern: A grep-like pattern to use to search for the Parquet
       files to process.
     output_file: The output file pattern to use when writing the basic blocks
       to disk.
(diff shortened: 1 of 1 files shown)


gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -109,6 +109,23 @@ def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
 def get_bbs(
```
ondrasej, line 109: Unrelated to this PR, and no action needed here. But we should see about turning this into a composite transform (per https://beam.apache.org/documentation/programming-guide/#composite-transforms).

Created #190 so that we don't forget about it.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -109,6 +109,23 @@ def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
...
     input_file_pattern: str, output_file: str
 ) -> Callable[[beam.Pipeline], None]:
+  """Creates a pipeline to process BBs from IR modules.
+
+  This function returns a function that builds a beam pipeline to automatically
+  load ir files from a ComPile style parquet file, process them into assembly
```
ondrasej, line 115: Very very nit:
- s/ir/IR/
- s/parquet/Parquet/

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -109,6 +109,23 @@ def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
...
+  basic blocks, deduplicate them, and then write them to a text file.
+
+  Args:
+    input_file_pattern: A grep-like pattern to use to search for the parquet
```
ondrasej, line 119: s/parquet/Parquet/

ondrasej (approved):

PR #186, opened 2024-09-03
Refactor utility to get llvm binary paths to common location

boomanaiden154:
This patch refactors utilities to get LLVM binary paths, like opt and llc, to a common location. This should make maintenance slightly easier. It also gives us a central location where we can override how these are found, to adapt for internal differences.


gematria/utils/python/runfiles.py
ondrasej, line 1: I'd prefer to put this somewhere under `gematria/llvm/...`. I don't think we want to add any other (non-LLVM) binaries to this library, because it would mean that everyone who uses one of them needs to carry over all others, even if they are never used.

Maybe we should have one library per tool even for the LLVM tools, but let's do that only when we need to use the tools separately.

ondrasej (changes requested):
Overall, this makes sense. But I'd prefer to make this a bit more specific (to avoid situations where we have one library that pulls in all the binaries used throughout the project, even if only one of them is used).

gematria/utils/python/runfiles.py
boomanaiden154: The binaries aren't getting added against the library. The idea is to keep them in the `data` tag of the binary/library actually using them, and just use this as a lookup function that we can easily override internally.

gematria/llvm/python/runfiles.py
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from rules_python.python.runfiles import runfiles
+
+
+def get_llvm_binary_path(tool_name: str) -> str:
```
ondrasej, line 20: Nit: Please add a docstring.

ondrasej (approved):

PR #187, opened 2024-09-03
Add filtering and processing BBs to the compilation pipeline

boomanaiden154:
This patch adds an additional step for filtering and processing basic blocks to the compilation pipeline. This gets us one step closer to fully annotating blocks and then saving them before another benchmarking pipeline takes over.

### DIFF
--- a/gematria/datasets/pipelines/compile_modules_lib.py
+++ b/gematria/datasets/pipelines/compile_modules_lib.py
@@ -122,10 +122,8 @@ class ProcessAndFilterBBs(beam.DoFn):
     output_block = self._bb_processor_filter.remove_risky_instructions(
         bb_hex, bb_hex, self._remove_memory_accessing_instructions
     )
-    if output_block == '':
-      return []
-
-    return [output_block]
+    if output_block != '':
+      yield output_block
 
 
 def get_bbs(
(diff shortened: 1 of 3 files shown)


gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -106,8 +107,31 @@ def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
...
+    self._bb_processor_filter = process_and_filter_bbs.BBProcessorFilter()
+
+  def process(self, bb_hex: str) -> Iterable[str]:
+    output_block = self._bb_processor_filter.remove_risky_instructions(
+        bb_hex, bb_hex, self._remove_memory_accessing_instructions
+    )
+    if output_block == '':
+      return []
+
+    return [output_block]
```
ondrasej, line 128: IIRC, it is more common for `process()` functions to `yield` than return a collection.

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -83,6 +83,28 @@ def test_deduplicate_bbs(self):
...
+    self.assertLen(filtered_bbs, 1)
+    self.assertEqual(filtered_bbs[0], 'B801000000')
+
+    # Check that we return an empty list if we have a BB that contains only
+    # instructions that should get filtered.
+
+    bb_hex = 'C3'
+
+    filtered_bbs = process_and_filter_transform.process(bb_hex)
+
```
ondrasej, line 105: Remove (some of) the empty lines?

ondrasej (approved):

PR #193, opened 2024-09-06
Minor cleanup in datasets

boomanaiden154:
This patch does some minor cleanup in gematria/datasets such as removing unused includes, adding includes where necessary to be IWYU compliant, and adding docstrings where necessary.

This based off of feedback from the 9-4-24 import.

### DIFF
--- a/gematria/datasets/annotating_importer_test.cc
+++ b/gematria/datasets/annotating_importer_test.cc
@@ -21,7 +21,6 @@
 #include <vector>
 
 #include "absl/status/statusor.h"
-#include "absl/strings/string_view.h"
 #include "gematria/llvm/canonicalizer.h"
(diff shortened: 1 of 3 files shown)


gematria/datasets/annotating_importer_test.cc
```diff
@@ -14,11 +14,14 @@
...
 #include "gematria/datasets/annotating_importer.h"
 
+#include <cassert>
 #include <memory>
 #include <string>
 #include <string_view>
 #include <vector>
 
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
```
ondrasej, line 24: I'd change `absl::string_view` -> `std::string_view` rather than add the import. We should aim for less Abseil in the code, not more :)

ondrasej (approved):

PR #194, opened 2024-09-06
Add docstrings for exegesis python bindings

boomanaiden154:
This patch adds docstrings for the ExegesisBenchmark python bindings so that they are better documented when working on them and also better documented when using them.


gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -67,22 +68,73 @@ PYBIND11_MODULE(exegesis_benchmark, m) {
...
+      .def(
+          "create",
+          []() -> absl::StatusOr<std::unique_ptr<ExegesisBenchmark>> {
+            InitializeForExegesisOnce();
+
+            return LlvmExpectedToStatusOr(ExegesisBenchmark::create());
+          },
+          R"(Creates a ExegesisBenchmark Instance.
+
+          Does the necessary initialization to run Exegesis and then creates
```
orodley, line 80: What happens if we call this twice -- does it do this initialisation again, or only the first time it's called?

gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -67,22 +68,73 @@ PYBIND11_MODULE(exegesis_benchmark, m) {
...
+
+          Takes a BenchmarkCode instance and then executes it, collecting
+          performance information at the same time.
+
+          Args:
+            input_benchmark: The BenchmarkCode instance formed from the block and
+              annotations of interest that should be benchmarked.
+          
+          Returns:
+            A floating point value represnting the inverse throughput of the block.
```
orodley, line 133: A bit more information on the units (IIUC this is a count of cycles taken to execute the block) would be useful here for interpreting the result

orodley, line 133: Typo: represnting

gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -67,22 +68,73 @@ PYBIND11_MODULE(exegesis_benchmark, m) {
...
+          "create",
+          []() -> absl::StatusOr<std::unique_ptr<ExegesisBenchmark>> {
+            InitializeForExegesisOnce();
+
+            return LlvmExpectedToStatusOr(ExegesisBenchmark::create());
+          },
+          R"(Creates a ExegesisBenchmark Instance.
+
+          Does the necessary initialization to run Exegesis and then creates
+          a ExegesisBenchmark that can then be used to execute annotated
```
orodley, line 81: Nit: a -> an (and below in the "Returns" section)

gematria/datasets/python/exegesis_benchmark.cc
orodley, line 80: Nit: I'd drop both the "then"s in this paragraph, as they're not necessary.

orodley (approved):

gematria/datasets/python/exegesis_benchmark.cc
boomanaiden154: Clarified.

gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -67,22 +68,76 @@ PYBIND11_MODULE(exegesis_benchmark, m) {
...
-        return LlvmExpectedToStatusOr(Self.benchmarkBasicBlock(InputBenchmark));
-      });
+      .def(
+          "create",
+          []() -> absl::StatusOr<std::unique_ptr<ExegesisBenchmark>> {
+            InitializeForExegesisOnce();
+
+            return LlvmExpectedToStatusOr(ExegesisBenchmark::create());
+          },
+          R"(Creates a ExegesisBenchmark Instance.
```
ondrasej, line 78: s/a/an/

gematria/datasets/python/exegesis_benchmark.cc
```diff
@@ -67,22 +68,76 @@ PYBIND11_MODULE(exegesis_benchmark, m) {
...
+
+          Does the necessary initialization to run Exegesis and creates a
```
ondrasej, line 80: ditto

ondrasej (approved):

Issue #201, opened 2024-09-10
Segmentation fault with cpp protos in compile_modules_lib


boomanaiden154:
When adding any protobuf library as a dependency to `compile_modules_lib`, we get a segmentation fault (at least in `compile_modules_lib_test`) rather than the expected behavior:

```
Fatal Python error: Segmentation fault

Thread 0x00007f3a20e00640 (most recent call first):
  File "/usr/lib/python3.10/threading.py", line 324 in wait
  File "/usr/lib/python3.10/threading.py", line 607 in wait
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/worker/data_plane.py", line 255 in run
  File "/usr/lib/python3.10/threading.py", line 1016 in _bootstrap_inner
  File "/usr/lib/python3.10/threading.py", line 973 in _bootstrap

Current thread 0x00007f3ad2b9d1c0 (most recent call first):
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/portability/fn_api_runner/fn_runner.py", line 1048 in _run_bundle
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/portability/fn_api_runner/fn_runner.py", line 811 in _execute_bundle
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/portability/fn_api_runner/fn_runner.py", line 483 in run_stages
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/portability/fn_api_runner/fn_runner.py", line 228 in run_via_runner_api
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/portability/fn_api_runner/fn_runner.py", line 204 in run_pipeline
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/runners/direct/direct_runner.py", line 128 in run_pipeline
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/pipeline.py", line 587 in run
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/pipeline.py", line 563 in run
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/testing/test_pipeline.py", line 115 in run
  File "/usr/local/lib/python3.10/dist-packages/apache_beam/pipeline.py", line 613 in __exit__
[...]

PR #202, opened 2024-09-10
Add support for annotating blocks to the pipeline

boomanaiden154:
This patch makes the module compilation Beam pipeline. This also changes the pipeline to output TFRecord files containing protos of annotated blocks rather than just hex values.


gematria/datasets/pipelines/BUILD.bazel
```diff
@@ -25,6 +29,12 @@ gematria_py_binary(
 gematria_py_test(
     name = "compile_modules_lib_test",
     srcs = ["compile_modules_lib_test.py"],
+    # TODO(boomanaiden154): We use the python implementation here as otherwise
+    # we get a segmentation fault. Eventually this should be debugged and
+    # fixed.
```
ondrasej, line 34: Please add a mention of https://github.com/google/gematria/issues/201 in here (so that we can search TODOs by issues).

ondrasej (approved):

PR #204, opened 2024-09-10
Fix loading of annotated blocks for tests.

virajbshah:
* Fixes incorrect logic for loading of annotated basic blocks for testing in `basic_blocks_with_throughput.py`.
 * Previously, we would silently load un-annotated blocks.

### DIFF
--- a/gematria/testing/python/basic_blocks_with_throughput.py
+++ b/gematria/testing/python/basic_blocks_with_throughput.py
@@ -36,8 +36,11 @@ _ANNOTATED_BASIC_BLOCK_RESOURCE_PATH = os.path.join(
     _ROOT_PATH,
     'gematria/testing/testdata/annotated_basic_blocks_with_throughput.pbtxt',
 )
-# Parsed basic block. An exception is thrown if the basic blocks do not parse.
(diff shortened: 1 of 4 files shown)


gematria/testing/python/basic_blocks_with_throughput.py
```diff
@@ -48,23 +51,35 @@
...
+        _ANNOTATED_BASIC_BLOCKS = text_format.Parse(
+            f.read(), throughput_pb2.BasicBlockWithThroughputListProto()
+        )
+    assert _ANNOTATED_BASIC_BLOCKS is not None
+    return _ANNOTATED_BASIC_BLOCKS
+  else:  # get_annotated_blocks == False
+    if _BASIC_BLOCKS is None:
+      runfiles_dir = os.environ.get('PYTHON_RUNFILES')
+      runfiles_env = runfiles.Create({'RUNFILES_DIR': runfiles_dir})
+      assert runfiles_env is not None
```
ondrasej, line 73: I'd move this out of the if/else statements (since it is the same for both cases).

ondrasej (approved):

ondrasej (approved):

Issue #200, opened 2024-09-09
Add layering check/header parsing to bazel flags


boomanaiden154:
Adding a `features = ["layering_check"]` to all our packages incrementally (and eventually somewhere common) would be ideal. It would prevent any transitive dependency issues and would also make imports into `google3` easier where this check is enabled by default.

boomanaiden154:
All of the core packages have this enabled now, closing.

PR #210, opened 2024-09-14
Enable layering_checks for gematria/utils

boomanaiden154:
This will more closely match the setup that we have internally and prevent transitive dependency issues.

Fixes part of #200.


ondrasej (approved):
Nice!

PR #228, opened 2024-09-28
Bump protobuf to most recent release

boomanaiden154:
This patch bumps our protobuf dependency to the most recent release while changing around the workspace file so that everything still works.


boomanaiden154:
Closing for now. Bumping to the latest version causes test failures in Sonnet, which is much more complicated in terms of actually upgrading.

PR #229, opened 2024-09-28
Add metrics for module compilation pipeline

boomanaiden154:
This patch adds metrics for the module compilation/annotation pipeline. These will show up in the Beam executor dashboard (i.e., GCP Dataflow, Spark, Flink, etc.) and provide valuable insight about what is actually happening, as currently a lot of the information is lost.


ondrasej (changes requested):
Nice! This will make it easier to see what's going on. I've added just a few naming nits, otherwise this looks good.

ondrasej (approved):

Issue #221, opened 2024-09-16
Handle additional syscall cases/terminator instructions


boomanaiden154:
There are other ways to make syscalls on X86, including `int 0x80`, and `sysenter` that we should probably also be sanitizing for security reasons.

We also need to make sure that we're taking care of all terminator instructions. We seem to be, but changing to `MCInstrDesc::isTerminator` probably makes a lot of sense and avoids us needing to manually handle the cases.

PR #231, opened 2024-09-28
Better syscall filtering

boomanaiden154:
This patch adds a couple more filters for various syscall types, which should be exhaustive for X86. Some of these are designed for 32-bit systems, so don't really work on 64-bit (where we will be benchmarking), but they might end up somewhere, and we still want to remove them.

Fixes part of #221.

### DIFF
--- a/gematria/datasets/process_and_filter_bbs_lib.cc
+++ b/gematria/datasets/process_and_filter_bbs_lib.cc
@@ -65,15 +65,39 @@ Expected<std::string> BBProcessorFilter::removeRiskyInstructions(
        *DisassembledInstructionsOrErr) {
     MCInstrDesc InstDesc =
         LLVMSupport->mc_instr_info().get(Instruction.mc_inst.getOpcode());
-    // TODO(boomanaiden154): This filtering is a bit simplistic currently. We
-    // should probably be using MCInsrtDesc::hasUnmodeledSideEffects, but this
-    // needs to be evaluated at scale.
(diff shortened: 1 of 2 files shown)


gematria/datasets/process_and_filter_bbs_lib.cc
```diff
@@ -65,15 +65,39 @@ Expected<std::string> BBProcessorFilter::removeRiskyInstructions(
...
+        continue;
+      case X86::SYSEXIT64:
+        continue;
+      case X86::SYSCALL:
+        continue;
+      case X86::SYSRET:
+        continue;
+      case X86::SYSRET64:
+        continue;
+    }
```
ondrasej, line 92: There's also `llvm::MCInstrDesc::hasUnmodeledSideEffects()` that might be an interesting flag here. IIRC, this covers also some instructions that are "benign" like `REP MOVS`, but in general it might be a flag that the instruction needs to be considered.

A conservative approach could be to trap all instructions that have unmodeled side effects, and then only allow ones that we know to be safe.

ondrasej (approved):

gematria/datasets/process_and_filter_bbs_lib.cc
boomanaiden154: Yeah, `hasUnmodeledSideEffects` was mentioned in a comment above. I need to actually look into what instructions are covered by it, because it seems like it might be a bit more broad than what we actually want.

I'll look into it soonish.

PR #233, opened 2024-09-30
Minor cleanups for find_accessed_addrs

orodley:
* Rename "before_block" and "after_block" to "prologue" and "epilogue"
* Use a typedef for the function type rather than stating it in the cast
* Fix a typo

### DIFF
--- a/gematria/datasets/block_wrapper.h
+++ b/gematria/datasets/block_wrapper.h
@@ -17,7 +17,7 @@
 
 #include <stdint.h>
 
-#include "third_party/absl/types/span.h"
+#include "absl/types/span.h"
 
 extern "C" {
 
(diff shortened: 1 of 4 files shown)


gematria/datasets/block_wrapper.h
```diff
@@ -17,33 +17,42 @@
 
 #include <stdint.h>
 
-#include "absl/types/span.h"
+#include "third_party/absl/types/span.h"
```
boomanaiden154, line 20: This change should probably be reveerted?

boomanaiden154 (approved):
LGTM.

Thanks for the cleanup around the before/after block language. it will hopefully keep things clearer when we get to adding block context to the model.

gematria/datasets/block_wrapper.h
orodley: Ah whoops, missed that one.

PR #234, opened 2024-09-30
Support vector registers in FindAccessedAddrs

orodley:
We only support 8 byte initial values, which are broadcast across the register. To avoid introducing unnecessary uarch dependencies, we find the largest vector register size used in the snippet, and only copy into that. Similarly, we detect whether the 16 extended registers introduced in AVX-512 are used, and only copy into them if so.

If no vector registers are used at all, we will skip over all instructions touching vector registers to avoid as much of the overhead as possible.


ondrasej (approved):

ondrasej:
LGTM for the code and the tests.

It looks like the test runners do not support ZMM instructions. We should perhaps add a check that the ZMM registers are supported and bail out if they are not (not in this PR though).

orodley:
> LGTM for the code and the tests.
> 
> It looks like the test runners do not support ZMM instructions. We should perhaps add a check that the ZMM registers are supported and bail out if they are not (not in this PR though).

I've added this check since it was pretty simple to do and keeps the tests green. I just added a small subclass test fixture which checks `cpuid` and skips the test if the AVX512 extensions we need aren't available.

PR #235, opened 2024-10-14
Switch runfiles usage to python library

boomanaiden154:
A while back, the canonical way within Gematria to get bazel runfiles was switched to using the runfiles library installed through pypi. This patch makes annotating_importer_test use this as it seems like this was written before the change and did not get updated afterwards.

### DIFF
--- a/gematria/datasets/python/BUILD.bazel
+++ b/gematria/datasets/python/BUILD.bazel
@@ -94,7 +94,6 @@ gematria_py_test(
         "//gematria/proto:basic_block_py_pb2",
         "//gematria/proto:canonicalized_instruction_py_pb2",
         "//gematria/proto:throughput_py_pb2",
-        "@rules_python//python/runfiles",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


virajbshah (commented):
Thanks for catching this!

gematria/datasets/python/BUILD.bazel
```diff
@@ -94,7 +94,6 @@ gematria_py_test(
         "//gematria/proto:basic_block_py_pb2",
         "//gematria/proto:canonicalized_instruction_py_pb2",
         "//gematria/proto:throughput_py_pb2",
-        "@rules_python//python/runfiles",
```
ondrasej, line 97: Is there anything we can drop from the `WORKSPACE` file?

ondrasej (approved):
Nice, thanks for cleaning this up!

gematria/datasets/python/BUILD.bazel
boomanaiden154: Yeah, looks like we can just remove the entire `rules_python` repo. I'll do a follow up patch to pull that out.

PR #236, opened 2024-10-14
Change unittest.TestCase to absl TestCase

boomanaiden154:
unittest.TestCase has no documented assertLen function. Change this to `absltest.TestCase` as the subclass so that the type-checker is happy even though this class is exclusively a mix-in.

### DIFF
--- a/gematria/testing/python/basic_blocks_with_throughput.py
+++ b/gematria/testing/python/basic_blocks_with_throughput.py
@@ -18,9 +18,9 @@ import copy
 import itertools
 import os
 
+from absl.testing import absltest
 from google.protobuf import text_format
 from runfiles import runfiles
-from absl.testing import absltest
 
 from gematria.basic_block.python import basic_block
 from gematria.basic_block.python import throughput
--- a/gematria/testing/python/basic_blocks_with_throughput.py
+++ b/gematria/testing/python/basic_blocks_with_throughput.py
@@ -189,8 +189,8 @@ class TestCase(unittest.TestCase):
     self.annotated_block_protos = get_basic_blocks(
         self.num_blocks, get_annotated_blocks=True, keep_fn=self.keep_function
     )
-    self.assertLen(self.block_protos, self.num_blocks)
-    self.assertLen(self.annotated_block_protos, self.num_blocks)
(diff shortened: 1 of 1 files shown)


ondrasej:
`assertLen` and a bunch of other assertion methods are provided by `absltest` (see [here](https://github.com/abseil/abseil-py/blob/f7ed4ac64809756e271e28ae413bc97825763cfd/absl/testing/absltest.py#L993)).

This class is used as a mixin to other tests, and I suspect that those are based on `absltest` and thus have this method. It still looks like a type checker failure though.

So, if something, I'd keep `assertLen` rather than `assertEqual`, and instead change the "base" class from `unittest.TestCase` to `absltest.TestCase`.

ondrasej (changes requested):
Marking this as "Request changes".

boomanaiden154:
> This class is used as a mixin to other tests, and I suspect that those are based on absltest and thus have this method. It still looks like a type checker failure though.

Yeah, definitely a type-checker failure.

> So, if something, I'd keep assertLen rather than assertEqual, and instead change the "base" class from unittest.TestCase to absltest.TestCase.

Fixed.

gematria/testing/python/basic_blocks_with_throughput.py
```diff
@@ -17,15 +17,16 @@
 import copy
 import itertools
 import os
-import unittest
+
+from google.protobuf import text_format
+from runfiles import runfiles
+from absl.testing import absltest
```
ondrasej, line 23: Please sort these by package name lexicographically.

ondrasej, line 23: We probably need to add this to the `BUILD` file, no?

ondrasej (changes requested):
One more minor comment, otherwise this looks good

gematria/testing/python/basic_blocks_with_throughput.py
boomanaiden154: Not on the open source side. It's a dependency installed through `pip`, so nothing in the `BUILD.bazel` is needed here.

ondrasej (approved):

PR #237, opened 2024-10-21
Replace cpuid usage with __builtin_cpu_supports

boomanaiden154:
This patch replaces use of cpuid in inline assembly with calls to __builtin_cpu_supports. This makes the code significantly simpler and also means that all of the footguns involved with cpuid are handled by the abstraction rather than manually.

### DIFF
--- a/gematria/datasets/find_accessed_addrs_test.cc
+++ b/gematria/datasets/find_accessed_addrs_test.cc
@@ -226,11 +226,8 @@ TEST_F(FindAccessedAddrsTest, YmmRegister) {
 class FindAccessedAddrsAvx512Test : public FindAccessedAddrsTest {
  protected:
   void SetUp() override {
-    int cpu_info;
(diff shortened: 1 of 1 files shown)


ondrasej (approved):
Nice, this is much more readable. I'm worried about the portability of the code, but it looks like both clang and gcc have this function. No MSVC though (but the current assembly code wouldn't work there either).

boomanaiden154:
> Nice, this is much more readable. I'm worried about the portability of the code, but it looks like both clang and gcc have this function. No MSVC though (but the current assembly code wouldn't work there either).

And `find_accessed_addrs` won't work on Windows anyways without significant work, so should probably be fine. Thanks for pointing this out though.

If we need cross-platform compatibility at some point in the future, we can probably use wrappers in `cpuid.h` like `__cpuidex`, which should be supported by all three as of clang 19.

PR #243, opened 2024-10-27
Add GetVocab DoFn in compile_modules_lib

boomanaiden154:
This patch adds an Apache Beam DoFn that gets the vocab from hex strings of basic blocks. This is intended to be used in the annotation pipeline to emit a vocab file along with the annotated blocks rather than having to manually construct it later without the benefit of all the parallelism.


gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -197,6 +201,26 @@ def process(
...
+  def process(self, bb_hex: str) -> Iterable[str]:
+    raw_block_proto = self._importer.basic_block_proto_from_hex(bb_hex)
+    block_proto = basic_block_protos.basic_block_from_proto(raw_block_proto)
+    tokens = set()
+
+    for instruction in block_proto.instructions:
+      tokens.update(instruction.as_token_list())
+
+    for token in tokens:
+      yield token
```
ondrasej, line 221: Shorter:
```py
yield from tokens
```

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -98,6 +98,38 @@ def test_annotate_bbs(self):
...
+            'EDX',
+            'RSP',
+            '_NO_REGISTER_',
+            'SAR',
+            'CMP',
+            'RDX',
+            'SUB',
+            '_DISPLACEMENT_',
+        ],
+    )
```
ondrasej, line 131: The expected list already contains 17 elements, i.e. it covers all tokens from the basic blocks.

Then, this is equivalent to:
```py
self.assertCountEqual(vocab, [...the list of expected tokens...])
```
but `assertCountEqual` will produce slightly better error messages on failure (it should print the actual list even if the number of items doesn't match).

ondrasej (approved):
LGTM overall, with two style nits.

PR #244, opened 2024-10-27
Add support for emitting vocab in the compile modules pipeline

boomanaiden154:
This patch adds emitting vocab directly to the module compilation/annotation pipeline so that vocab can be extracted in parallel to annotation.


gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -190,6 +192,16 @@ def test_get_bbs(self):
...
 
+    with open(
+        vocab_output_file_pattern + '-00000-of-00001'
+    ) as vocab_file_handle:
+      vocab_tokens = [token.strip() for token in vocab_file_handle.readlines()]
+
+    self.assertLen(vocab_tokens, 4)
+    self.assertContainsSubset(
+        ['_D_', '_IMMEDIATE_', 'MOV', 'EAX'], vocab_tokens
+    )
```
ondrasej, line 203: `self.assertCountEqual()` for a single assertion that does both (and has better error messages).

PR #248, opened 2024-10-28
Remove `owning_binary_` from `AnnotatingImporter` state.

virajbshah:
* There is no longer any point of keeping `owning_binary_` as part of the `AnnotatingImporter` state, since it is not re-used between calls to `GetAnnotatedBasicBlockProtos`, which is the intended use of `AnnotatingImporter`.
 * This change is meant to help merge some functionality from `annotating_importer` with `extract_bbs_from_obj_lib`.

### DIFF
--- a/gematria/datasets/annotating_importer.cc
+++ b/gematria/datasets/annotating_importer.cc
@@ -89,27 +89,26 @@ absl::Status AnnotatingImporter::LoadPerfData(std::string_view file_name) {
   return absl::OkStatus();
 }
 
-absl::Status AnnotatingImporter::LoadBinary(std::string_view file_name) {
+absl::StatusOr<llvm::object::OwningBinary<llvm::object::Binary>>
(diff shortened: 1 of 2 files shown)


gematria/datasets/annotating_importer.cc
```diff
@@ -365,19 +357,25 @@ absl::StatusOr<std::vector<BasicBlockWithThroughputProto>>
 AnnotatingImporter::GetAnnotatedBasicBlockProtos(
     std::string_view elf_file_name, std::string_view perf_data_file_name,
     std::string_view source_name) {
-  absl::Status status = LoadBinary(elf_file_name);
-  if (!status.ok()) {
-    return status;
+  auto owning_binary = LoadBinary(elf_file_name);
```
boomanaiden154, line 360: Prefer an explicit type here since it isn't immediately obvious.

boomanaiden154 (approved):
Seems reasonable enough to me.

PR #249, opened 2024-10-30
Fix use after move to get exegesis annotator working in pipeline

boomanaiden154:
This patch fixes a user after move in BHiveToExegesis/related classes where we would create a LLVMState, pass it into a couple other constructors to create things that would store a reference to LLVMState, and then promptly move it to call the BHiveToExegesis constructor. This resulted in use after moves, which were ultimately caught by ubsan. This patch fixes that behavior by encapsulating LLVMState within a unique_ptr to prevent the problems.

The exegesis annotator is also enabled in the compile_modules invocation script and test coverage is added now that it works.

### DIFF
--- a/gematria/datasets/bhive_to_exegesis.cc
+++ b/gematria/datasets/bhive_to_exegesis.cc
@@ -47,7 +47,7 @@ namespace gematria {
 
 BHiveToExegesis::BHiveToExegesis(
     LlvmArchitectureSupport& ArchitectureSupport,
-    llvm::exegesis::LLVMState&& LLVMExegesisState,
+    std::unique_ptr<llvm::exegesis::LLVMState>&& LLVMExegesisState,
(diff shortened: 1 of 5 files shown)


gematria/datasets/bhive_to_exegesis.cc
```diff
@@ -47,7 +47,7 @@ namespace gematria {
 
 BHiveToExegesis::BHiveToExegesis(
     LlvmArchitectureSupport& ArchitectureSupport,
-    llvm::exegesis::LLVMState&& LLVMExegesisState,
+    std::unique_ptr<llvm::exegesis::LLVMState>&& LLVMExegesisState,
```
orodley, line 50: I think passing the `unique_ptr` by value instead of rvalue reference is a bit clearer and more standard. Understood that this is matching existing usage though, so no need to block the PR over this.

orodley (approved):

Issue #251, opened 2024-10-30
Exegesis annotator failing in machine code verification


boomanaiden154:
The exegesis annotator is currently failing on at least one block with the following machine code verification error.

```
*** Bad machine code: Using an undefined physical register ***
 2024-10-29 23:53:48.280841-07000   - function:    foo
 2024-10-29 23:53:48.280848-07000   - basic block: %bb.0  (0x7fe694a284c8)
 2024-10-29 23:53:48.280959-07000   - instruction: MOVSQ $rdi, $rsi, $noreg, implicit-def $edi, implicit-def $esi, implicit $edi, implicit $esi, implicit $df
 2024-10-29 23:53:48.281092-07000   - operand 7:   implicit $df
```

It looks like this needs to get fixed during the annotation stage.

boomanaiden154:
```patch
diff --git a/gematria/datasets/basic_block_utils.cc b/gematria/datasets/basic_block_utils.cc
--- a/gematria/datasets/basic_block_utils.cc
+++ b/gematria/datasets/basic_block_utils.cc
@@ -33,6 +33,12 @@ namespace gematria {
 
 unsigned getSuperRegister(unsigned OriginalRegister,
                           const MCRegisterInfo &RegisterInfo) {
+  // Sections of EFLAGS that are explicitly modeled by LLVM as separate
+  // registers should all return eflags.
+  if (OriginalRegister == X86::EFLAGS || OriginalRegister == X86::DF) {
+    return X86::EFLAGS;
+  }
+
   // We should be able to get the super register by getting an iterator from
   // RegisterInfo and getting the back, but this always returns a value of
   // zero.
diff --git a/gematria/datasets/basic_block_utils_test.cc b/gematria/datasets/basic_block_utils_test.cc
--- a/gematria/datasets/basic_block_utils_test.cc
+++ b/gematria/datasets/basic_block_utils_test.cc
@@ -189,6 +189,14 @@ TEST_F(BasicBlockUtilsTest, DISABLED_Use
               UnorderedElementsAre(X86::RAX, X86::RDX, X86::RCX));
 }
 
+TEST_F(BasicBlockUtilsTest, MovsqImplicitDfUsesEflags) {
+  std::vector<unsigned> UsedRegisters = getUsedRegs(R"asm(
+    movsq
+  )asm");
+  EXPECT_THAT(UsedRegisters,
+              UnorderedElementsAre(X86::RSI, X86::RDI, X86::EFLAGS));
+}
+
[...]

PR #253, opened 2024-10-30
Add support for df register in basic_block_utils

boomanaiden154:
Certain instructions implicitly use the df register, which means we need to support it for the exegesis annotator so that we do not get MC verification errors. Additionally, it is good to explicitly define the state to ensure we have deterministic execution. This patch adds support for the df register by explicitly returning eflags as the super register as LLVM does not seem to explicitly model df as a subregister of eflags.

Part of a fix for #251.

### DIFF
--- a/gematria/datasets/basic_block_utils.cc
+++ b/gematria/datasets/basic_block_utils.cc
@@ -33,6 +33,12 @@ namespace gematria {
 
 unsigned getSuperRegister(unsigned OriginalRegister,
                           const MCRegisterInfo &RegisterInfo) {
+  // Sections of EFLAGS that are explicitly modeled by LLVM as separate
+  // registers should all return eflags.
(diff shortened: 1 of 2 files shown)


ondrasej (approved):
LGTM for this change.

Though eventually, we might want to do the same, and model them separately if they are truly independent of the rest of the EFLAGS register (e.g. it is executed independently on the silicon).

boomanaiden154:
> Though eventually, we might want to do the same, and model them separately if they are truly independent of the rest of the EFLAGS register (e.g. it is executed independently on the silicon).

Yes, that's something that I looked into. We would need support on the Exegesis side though, and they are part of the same physical register, so this works fine. Something I should probably aim to support in the future though, but then we run into issues like conflicting EFLAGS/DF values.

PR #254, opened 2024-10-30
Add support for instructions using more registers in exegesis annotator

### LINKED ISSUE #251: Exegesis annotator failing in machine code verification

boomanaiden154:
This patch adds explicit support in the exegesis annotator for instructions that use additional registers, after some minor refactoring to make things a bit more clear.

Fixes #251.

### DIFF
--- a/gematria/datasets/find_accessed_addrs_exegesis.cc
+++ b/gematria/datasets/find_accessed_addrs_exegesis.cc
@@ -142,6 +142,14 @@ Expected<ExecutionAnnotations> ExegesisAnnotator::findAccessedAddrs(
       BenchCode.Key.MemoryMappings.size());
 
   for (unsigned RegisterIndex : UsedRegisters) {
+    // Skip all the segment registers other than FS and GS as we cannot set
+    // them from userspace and the machine code verifier does not complain if
+    // they are not defined.
+    if (RegisterIndex == X86::SS || RegisterIndex == X86::CS ||
+        RegisterIndex == X86::DS || RegisterIndex == X86::ES) {
+      continue;
+    }
+
     RegisterAndValue *NewRegisterValue = MemAnnotations.add_initial_registers();
     NewRegisterValue->set_register_name(
         State.getRegInfo().getName(RegisterIndex));
@@ -157,7 +165,11 @@ Expected<ExecutionAnnotations> ExegesisAnnotator::findAccessedAddrs(
         MRI.getRegClass(X86::VK64RegClassID).contains(RegisterIndex)) {
       RegVal.Value = APInt(64, kInitialRegVal);
       NewRegisterValue->set_register_value(kInitialRegVal);
-    } else if (RegisterIndex == X86::EFLAGS || RegisterIndex == X86::MXCSR) {
+    } else if (RegisterIndex == X86::FS || RegisterIndex == X86::GS) {
+      RegVal.Value = APInt(32, kInitialRegVal);
+      NewRegisterValue->set_register_value(kInitialRegVal);
+    } else if (RegisterIndex == X86::EFLAGS || RegisterIndex == X86::MXCSR ||
+               RegisterIndex == X86::FPCW) {
       RegVal.Value = APInt(32, 0);
       NewRegisterValue->set_register_value(0);
     } else {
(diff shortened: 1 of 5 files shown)


gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -134,21 +135,30 @@ Expected<ExecutionAnnotations> ExegesisAnnotator::findAccessedAddrs(
   BenchCode.Key.MemoryValues["memdef1"] = MemVal;
 
   const llvm::MCRegisterInfo &MRI = State.getRegInfo();
+  std::vector<unsigned> UsedRegisters = gematria::getUsedRegisters(
```
ondrasej, line 138:
```cc
const std::vector<unsigned>
```

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -211,19 +221,8 @@ Expected<ExecutionAnnotations> ExegesisAnnotator::findAccessedAddrs(
     MemAnnotations.add_accessed_blocks(Mapping.Address);
   }
 
-  std::vector<unsigned> UsedRegisters = gematria::getUsedRegisters(
-      *DisInstructions, State.getRegInfo(), State.getInstrInfo());
-
   MemAnnotations.mutable_initial_registers()->Reserve(UsedRegisters.size());
```
ondrasej, line 224: This line should move along with the rest of the code (it's pre-allocating memory for all the `add_initial_regisers()` calls that were in the for loop below.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -134,21 +135,30 @@ Expected<ExecutionAnnotations> ExegesisAnnotator::findAccessedAddrs(
...
+
+    if (MRI.getRegClass(X86::GR64_NOREX2RegClassID).contains(RegisterIndex)) {
+      RegVal.Value = APInt(64, kInitialRegVal);
+      NewRegisterValue->set_register_value(kInitialRegVal);
+    } else if (MRI.getRegClass(X86::VR128RegClassID).contains(RegisterIndex)) {
+      RegVal.Value = APInt(128, kInitialRegVal);
+      NewRegisterValue->set_register_value(kInitialRegVal);
+    } else if (RegisterIndex == X86::EFLAGS) {
+      RegVal.Value = 0;
+      NewRegisterValue->set_register_value(0);
```
ondrasej, line 157: Some bits in `EFLAGS` can be set explicitly using separate instructions (e.g. the `DF` flag in question is modified using instructions `STD` and `CLD`). Similar for other flags.

I don't think we need to do anything about it now, but we should keep this in mind, and we might want to set it/allow setting it in the future.

ondrasej (changes requested):

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Yep. Support would need to be added on the Exegesis side for that. I can look at doing that at some point.

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -134,21 +136,49 @@ Expected<ExecutionAnnotations> ExegesisAnnotator::findAccessedAddrs(
...
-    RegisterValue RegVal;
-    RegVal.Register =
-        MRI.getRegClass(X86::GR64_NOREX2RegClassID).getRegister(i);
-    RegVal.Value = APInt(64, kInitialRegVal);
-    BenchCode.Key.RegisterInitialValues.push_back(RegVal);
-  }
+  for (unsigned RegisterIndex : UsedRegisters) {
+    // Skip all the segment registers other than FS and GS as we cannot set
+    // them from userspace and the machine code verifier does not complain if
+    // they are not defined.
```
ondrasej, line 147: Is there a register class covering the segment registers? There might even be one covering the user-assignable registers `FS` and `GS` or the hard-coded ones. This is an architecture-defined thing (not LLVM-defined or ABI-defined) so there might very likely be one.

ondrasej (approved):

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: There's one for the segment registers, but not one that explicitly excludes FS/GS.

PR #257, opened 2024-11-03
Add support for annotating FP registers

boomanaiden154:
This patch adds support for annotating FP registers, including support the LLVM FP pseudo registers which seem to crop up from time to time in the MC layer (although I do not think I can write tests for them).

### DIFF
--- a/gematria/datasets/find_accessed_addrs_exegesis.cc
+++ b/gematria/datasets/find_accessed_addrs_exegesis.cc
@@ -32,6 +32,7 @@
 #include "gematria/proto/execution_annotation.pb.h"
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/bit.h"
 #include "llvm/MC/MCContext.h"
 #include "llvm/MC/MCInst.h"
 #include "llvm/MC/MCRegisterInfo.h"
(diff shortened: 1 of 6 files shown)


orodley (approved):

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -61,6 +61,12 @@ using namespace llvm::exegesis;
 constexpr uint64_t kInitialRegVal = 0x12345600;
 constexpr uint64_t kInitialMemVal = 0x12345600;
 
+// TODO(boomanaiden154): This is currently chosen just to make things simple.
+// Eventually we should probably switch to a different constant per FP
+// FP register size and ensure it is a reasonable enough value to hopefully
```
ondrasej, line 66: Remove one `FP` (on this line or the following line).

gematria/datasets/find_accessed_addrs_exegesis.cc
```diff
@@ -61,6 +61,12 @@ using namespace llvm::exegesis;
...
+// avoid problems like expensive FP operations due to edge cases.
+constexpr uint64_t kInitialFloatVal = 0x12345600;
```
ondrasej, line 68: Does this produce a reasonable floating point value? (We might want to do a `bit_cast` from a double just to be safe).

ondrasej (changes requested):

gematria/datasets/find_accessed_addrs_exegesis.cc
boomanaiden154: Used `llvm::bit_cast` to get a reasonable value (10^6) for floats and doubles. Will need some more thought on how exactly we do this in the future, so left the comment (modified).

This is mostly just aimed at a stop-gap so we can avoid MC verification errors when annotating.

ondrasej (approved):

PR #261, opened 2024-11-18
Add APX register support

orodley:


gematria/datasets/find_accessed_addrs.h
```diff
@@ -39,26 +39,40 @@ enum class VectorRegWidth : uint64_t {
...
+  //
+  // If this is false, then the latter 16 elements of vector_regs are unset and
+  // should be ignored.
+  uint64_t uses_upper_vector_regs;  // offset 0x8
+  // If true, the code uses at least one of the 16 extra general purpose
+  // registers defined in APX.
+  //
+  // If this is false, then the elements of apx_regs are unset and should be
+  // ignored.
+  uint64_t uses_apx_regs;  // offset 0x10
```
ondrasej, line 56: As discussed internally: Let's add a note that this is `int64_t` and not `bool` to make the layout of the struct simpler.

ondrasej (approved):

PR #266, opened 2024-12-18
Add extra checks to select the right mapping and filter samples within it.

virajbshah:
* Add a binary name check while identifying the main mapping.
 * A new check of the sample PID against the main mapping PID.
 * Remove the check for whether the block's code was within the bounds
   of the executable segment in terms of file addresses, this is
   redundant as long as the mapping is valid.

### DIFF
--- a/gematria/datasets/annotating_importer.cc
+++ b/gematria/datasets/annotating_importer.cc
@@ -267,6 +267,9 @@ AnnotatingImporter::GetSamples(
     }
 
     // Filter out sample events from outside the profiled binary.
+    if (!event.sample_event().has_pid() ||
+        !(event.sample_event().pid() == mapping->pid()))
+      continue;
     uint64_t sample_ip = event.sample_event().ip();
     if (sample_ip < mmap_begin_addr || sample_ip >= mmap_end_addr) continue;
 
@@ -329,6 +332,13 @@ AnnotatingImporter::GetLBRBlocksWithLatency(
         !event.sample_event().branch_stack_size()) {
       continue;
     }
+
+    // Check if the sample PID matches that of the relevant mapping.
+    if (!event.sample_event().has_pid() ||
+        !(event.sample_event().pid() == mapping->pid())) {
+      continue;
+    }
+
     const auto &branch_stack = event.sample_event().branch_stack();
     for (int branch_idx = branch_stack.size() - 2; branch_idx >= 0;
          --branch_idx) {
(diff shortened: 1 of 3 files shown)


gematria/datasets/annotating_importer.cc
```diff
@@ -243,13 +260,16 @@ AnnotatingImporter::GetSamples() {
...
+  for (const auto &event : perf_data->events()) {
     // Filter out non-sample events.
     if (!event.has_sample_event()) {
       continue;
     }
 
     // Filter out sample events from outside the profiled binary.
+    if (!event.sample_event().has_pid() ||
+        !(event.sample_event().pid() == mapping->pid()))
+      continue;
```
ondrasej, line 272: Very nit: Please put braces around this statement, for consistency with the rest of the code.

gematria/datasets/annotating_importer.cc
```diff
@@ -73,17 +73,26 @@ absl::StatusOr<const quipper::PerfDataProto *> AnnotatingImporter::LoadPerfData(
   return &perf_reader_.proto();
 }
 
+namespace {
+
+std::string GetFileNameFromPath(const std::string &path) {
```
ondrasej, line 78: Nit: This is usually called "Basename" in the unix world.

ondrasej (changes requested):

gematria/datasets/annotating_importer.cc
```diff
@@ -49,6 +49,11 @@
 
 namespace gematria {
 
+// Memory mapping protection flag bits on Linux, from `sys/mman.h`.
+constexpr int kProtRead = 0b001;  /* PROT_READ */
+constexpr int kProtWrite = 0b010; /* PROT_WRITE */
+constexpr int kProtExec = 0b100;  /* PROT_EXEC */
```
ondrasej, line 55: Can we use `PROT_READ`, `PROT_WRITE`, and `PROT_EXEC` directly in our code?

Normally, they should be provided by `sys/mman.h` - and if they are, I'd go for the system definitions rather than define our own.

ondrasej (approved):

gematria/datasets/annotating_importer.cc
virajbshah: They are indeed provided by `sys/mman.h`, but I opted to define them here with the comment pointing to `sys/mman.h` so that it doesn't stop this from being built on Windows. I _could_ use a `__has_include(<sys/mman.h>)` with a preprocessor macro at the top to use `sys/mman.h` whenever possible and just define `PROT_{READ,WRITE,EXEC}` otherwise. Or would simply dropping Windows compatibility make more sense?

ondrasej: We already have a Linux-only (POSIX maybe) dependency with finding accessed addresses, so I'm not too worried about that.

But we could do the `__has_include` check and include it if defined, otherwise take our own definitions.

PR #267, opened 2024-12-18
Fix annotating importer address resolution.

virajbshah:
* Previously, the annotating importer would break in certain
   situations, for example if the virtual address the executable segment
   is to be loaded at was not page-aligned.


ondrasej (commented):
LGTM, once the comments in the parent PRs are resolved.

ondrasej (approved):

PR #270, opened 2024-12-25
Add basic benchmark scheduler

boomanaiden154:
This patch introduces the BenchmarkScheduler interface that sets the CPU
affinity mask for the current process and gets a benchmark core for use
by llvm-exegesis. This patch additional introduces a BenchmarkScheduler
implementation that simply does nothing, which will preserve the
behavior that we have currently where we do not explicitly set CPU
masks.

### DIFF
--- a/gematria/datasets/pipelines/benchmark_cpu_scheduler.py
+++ b/gematria/datasets/pipelines/benchmark_cpu_scheduler.py
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 import abc
-import typing_extensions
+from typing_extensions import override
+from collections.abc import Iterable
+import os
+import re
 
 
 class BenchmarkScheduler(metaclass=abc.ABCMeta):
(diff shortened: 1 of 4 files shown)


gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -0,0 +1,35 @@
...
+  @abstractmethod
+  def setup_and_get_benchmark_core(self) -> int | None:
+    pass
+
+  @abstractmethod
+  def verify(self):
+    pass
+
+
+class NoSchedulingBenchmarkScheduler(BenchmarkScheduler):
```
ondrasej, line 29: Please add a class docstring explaining what this scheduler does.

Docstrings for overridden methods are not strictly needed, but please add an `@override` decorator (from `typing_extensions` and not `typing` or compatibility with Python 3.11).

ondrasej (changes requested):

gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -12,24 +12,57 @@
...
 class NoSchedulingBenchmarkScheduler(BenchmarkScheduler):
+  """A basic BenchmarkScheduler implementation that does nothing.
+
+  This BenchmarkScheduler implementation does nothing. It leaves scheduling
+  of the benchmarking process to the operating system by specifying any core
+  can be used for benchmarking, performs no setup in the parent process, and
+  performs no verification.
+  """
 
+  @typing_extensions.override
```
ondrasej, line 62: Optional: `typing_extensions` (and also `collections.abc` and `typing`) has an exception from the "no direct imports" rule. But this is OK too.

ondrasej (approved):

PR #271, opened 2024-12-25
Add default benchmark scheduler

boomanaiden154:
This patch adds a default benchmark scheduler. This scheduler assumes
the worker performing the benchmarking has three hyperthreads available
inside its CPU mask, with two of them being part of a pair (from the
same core). It then selects one core from the pair to be used for the
benchmarking subprocess, and selects the core outside of the pair to be
used to run the main exegesis/python program. Verification is also added
to ensure that nothing external to the program changes the process mask,
or at least to catch it early if it does happen.

### DIFF
--- a/gematria/datasets/pipelines/benchmark_cpu_scheduler.py
+++ b/gematria/datasets/pipelines/benchmark_cpu_scheduler.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import abc
-import typing_extensions
+from typing_extensions import override
 from collections.abc import Iterable
 import os
 import re
@@ -62,11 +62,11 @@ class NoSchedulingBenchmarkScheduler(BenchmarkScheduler):
   performs no verification.
   """
 
-  @typing_extensions.override
+  @override
   def setup_and_get_benchmark_core(self) -> int | None:
     return None
 
-  @typing_extensions.override
+  @override
   def verify(self):
     pass
 
@@ -121,7 +121,7 @@ class DefaultBenchmarkScheduler(BenchmarkScheduler):
         'Expected a pair of neighboring hyperthreads in the CPU mask.'
     )
 
-  @typing_extensions.override
+  @override
   def setup_and_get_benchmark_core(self) -> int | None:
     cpu_mask = os.sched_getaffinity(0)
 
@@ -136,7 +136,7 @@ class DefaultBenchmarkScheduler(BenchmarkScheduler):
 
     return hyperthread_pair[0]
 
-  @typing_extensions.override
+  @override
   def verify(self):
     cpu_mask = list(os.sched_getaffinity(0))
     if self._cpu_mask != cpu_mask:
--- a/gematria/datasets/pipelines/benchmark_cpu_scheduler.py
+++ b/gematria/datasets/pipelines/benchmark_cpu_scheduler.py
@@ -14,7 +14,7 @@
 
 import abc
 from typing_extensions import override
-from collections.abc import Iterable
+from collections.abc import Collection
 import os
 import re
 
@@ -71,6 +71,19 @@ class NoSchedulingBenchmarkScheduler(BenchmarkScheduler):
     pass
 
 
+def _get_neighboring_threads(cpu_index: int) -> list[int]:
+  with open(
+      f'/sys/devices/system/cpu/cpu{cpu_index}/topology/thread_siblings_list'
+  ) as thread_sibling_list_handle:
+    neighboring_threads_strings = re.split(
+        '[-,]+', thread_sibling_list_handle.read().strip()
+    )
+    neighboring_threads = [
+        int(cpu_index_str) for cpu_index_str in neighboring_threads_strings
+    ]
+  return neighboring_threads
+
+
 class DefaultBenchmarkScheduler(BenchmarkScheduler):
   """A BenchmarkScheduler that schedules processes separately.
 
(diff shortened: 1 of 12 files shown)


gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -33,3 +36,62 @@ def setup_and_get_benchmark_core(self) -> int | None:
...
+      if (
+          neighboring_threads[0] in cpu_mask
+          and neighboring_threads[1] in cpu_mask
+      ):
+        cpus = list(cpu_mask)
+        cpus.remove(neighboring_threads[0])
+        cpus.remove(neighboring_threads[1])
+        return (cpus[0], [neighboring_threads[0], neighboring_threads[1]])
+    raise ValueError(
+        'Expected a pair of neighboring hyperthreads in the CPU mask.'
```
ondrasej, line 121: Nit: We need a pair of hyperthreads _and_ an aux thread (i.e. three threads total).

gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -33,3 +36,62 @@ def setup_and_get_benchmark_core(self) -> int | None:
...
+        raise ValueError('Expected two hyperthreads per CPU.')
+
+      if (
+          neighboring_threads[0] in cpu_mask
+          and neighboring_threads[1] in cpu_mask
+      ):
+        cpus = list(cpu_mask)
+        cpus.remove(neighboring_threads[0])
+        cpus.remove(neighboring_threads[1])
+        return (cpus[0], [neighboring_threads[0], neighboring_threads[1]])
```
ondrasej, line 119: I'd add one more check that `cpus` is non-empty after removing the hyperthread pair. Otherwise, this will raise a `IndexError` when there's just the hyperthread pair but no aux thread.

That said, it's probably better to invert the ifs to avoid nesting, e.g.
```py
if (
    neighboring_threads[0] not in cpu_mask
    or neighboring_threads[1] not in cpu_mask
):
  continue
cpus = list(cpu_mask)
cpus.remove(...)
cpus.remove(...)
if not cpus:
  continue
return (cpus[0], [...])
```

gematria/datasets/pipelines/benchmark_cpu_scheduler_test.py
```diff
@@ -24,6 +27,102 @@ def test_no_scheduling(self):
...
+    self.assertEqual(aux_cpu, expected_aux_cpu)
+    self.assertContainsSubsequence(hyperthread_pair, expected_hyperthread_pair)
+    self._reset_cpu_affinity(old_cpu_mask)
+
+  def test_default_scheduler_get_cores_no_neighboring_threads(self):
+    cpu_mask = os.sched_getaffinity(0)
+    three_cores = [cpu_mask.pop(), cpu_mask.pop(), cpu_mask.pop()]
+
+    scheduler = benchmark_cpu_scheduler.DefaultBenchmarkScheduler()
+    with self.assertRaises(ValueError):
```
ondrasej, line 76: `assertRaisesRegex` might be helpful here, to be sure that we get the expected error.

gematria/datasets/pipelines/benchmark_cpu_scheduler_test.py
```diff
@@ -24,6 +27,102 @@ def test_no_scheduling(self):
...
+    expected_aux_cpu, expected_hyperthread_pair, old_cpu_mask = (
+        self._set_normal_affinity()
+    )
+
+    scheduler = benchmark_cpu_scheduler.DefaultBenchmarkScheduler()
+    benchmark_core = scheduler.setup_and_get_benchmark_core()
+    self.assertIn(benchmark_core, expected_hyperthread_pair)
+    set_cpu_mask = os.sched_getaffinity(0)
+    self.assertLen(set_cpu_mask, 1)
+    self.assertEqual(set_cpu_mask.pop(), expected_aux_cpu)
```
ondrasej, line 89: Nit: `self.assertSequenceEqual(set_cpu_mask, (expected_aux_cpu,))` should do both in one statement and produce better error messages.

ondrasej (changes requested):

gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -33,3 +36,62 @@ def setup_and_get_benchmark_core(self) -> int | None:
...
+class DefaultBenchmarkScheduler(BenchmarkScheduler):
+
+  def __init__(self):
+    self._cpu_mask = []
+
+  @staticmethod
+  def _get_neighboring_threads(cpu_index: int) -> list[int]:
+    with open(
+        f'/sys/devices/system/cpu/cpu{cpu_index}/topology/thread_siblings_list'
+    ) as thread_sibling_list_handle:
```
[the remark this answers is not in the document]
boomanaiden154: This does not need root access, and shouldn't need any special permissions. I haven't looked too deeply into how sysfs permissions are setup, but this has worked everywhere I have tried it (local machine, container, GHA, internally in a beam pipeline).

gematria/datasets/pipelines/benchmark_cpu_scheduler_test.py
```diff
@@ -24,6 +27,102 @@ def test_no_scheduling(self):
...
+        self._set_normal_affinity()
+    )
+    scheduler = benchmark_cpu_scheduler.DefaultBenchmarkScheduler()
+    cpu_mask = os.sched_getaffinity(0)
+    aux_cpu, hyperthread_pair = scheduler._get_aux_core_and_hyperthread_pair(
+        cpu_mask
+    )
+    self.assertEqual(aux_cpu, expected_aux_cpu)
+    self.assertContainsSubsequence(hyperthread_pair, expected_hyperthread_pair)
+    self._reset_cpu_affinity(old_cpu_mask)
```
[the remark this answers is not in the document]
boomanaiden154: I set it up with `addCleanup` as we don't use this function for every test depending upon the setup that we need.

gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -70,45 +71,64 @@ def verify(self):
...
+  """A BenchmarkScheduler that schedules processes separately.
+
+  DefaultBenchmarkScheduler schedules the main process and the benchmark
+  subprocess on separate cores, making sure to reserve the second hyperthread
+  on the benchmarking core to prevent interference. It expects that the main
+  process is initially given a CPU Mask with three active threads, additionally
+  assuming that two of the threads are neighboring (part of the same core).
+  Errors are raised if these conditions are not met. The benchmarking core
+  returned is one of the two neighboring threads. The main process has its
+  COU mask limited to the thread that neighbors neither of the other threads.
```
ondrasej, line 97: s/COU/CPU/

gematria/datasets/pipelines/benchmark_cpu_scheduler_test.py
```diff
@@ -22,73 +22,83 @@
 
 class BenchmarkSchedulerTests(absltest.TestCase):
 
+  def _set_normal_affinity(self):
+    cpu_mask = os.sched_getaffinity(0)
+    cpu_mask_list = list(cpu_mask)
+
+    self.addCleanup(lambda: os.sched_setaffinity(0, cpu_mask_list))
```
ondrasej, line 29: I'd rather do `list(cpu_mask_list)` here, so that we always use a clean copy of the original mask when restoring the original state.

We return `cpu_mask_list` to the caller who is free to modify it any way they like.

ondrasej (changes requested):

ondrasej (approved):

gematria/datasets/pipelines/benchmark_cpu_scheduler_test.py
```diff
@@ -12,18 +12,123 @@
...
 from gematria.datasets.pipelines import benchmark_cpu_scheduler
 
 
 class BenchmarkSchedulerTests(absltest.TestCase):
 
+  def _set_normal_affinity(self):
+    cpu_mask = os.sched_getaffinity(0)
+    cpu_mask_list = list(cpu_mask)
+
+    self.addCleanup(lambda: os.sched_setaffinity(0, list(cpu_mask_list)))
```
ondrasej, line 29: Actually, this doesn't help because the `list()` statement inside the lambda will run at the cleanup, and `cpu_mask_list` has already been changed.

What we need is more like
```py
original_cpu_mask = list(cpu_mask_list)
self.addCleanup(lambda: os.sched_setaffinity(0, original_cpu_mask))
```

ondrasej (changes requested):

ondrasej (approved):

PR #272, opened 2024-12-30
Plumb pinned core through benchmarking abstractions

boomanaiden154:
This patch does the necessary plumbing to enable setting the
benchmarking core from the Python wrapper around exegesis that is used
for benchmarking.


gematria/datasets/exegesis_benchmark_lib_test.cc
ondrasej, line 1: We should also have a test that sets the CPU affinity, at least to see that it doesn't crash.

ondrasej (approved):

PR #273, opened 2024-12-30
Allow use of benchmark scheduler within benchmarking pipeline

boomanaiden154:
This patch adds support for using the newly added benchmark scheduler
classes within the benchmarking pipeline to assign a core to be used for
benchmarking.

### DIFF
--- a/gematria/datasets/pipelines/benchmark_bbs.py
+++ b/gematria/datasets/pipelines/benchmark_bbs.py
@@ -31,13 +31,10 @@ _INPUT_FILE_PATTERN = flags.DEFINE_string(
 _OUTPUT_FILE_PATTERN = flags.DEFINE_string(
     'output_file_pattern', None, 'The output file path/pattern.', required=True
 )
-_BENCHMARK_SCHEDULER = flags.DEFINE_enum(
+_BENCHMARK_SCHEDULER = flags.DEFINE_enum_class(
     'benchmark_scheduler',
-    'NoScheduling',
-    [
-        scheduler_type.name
-        for scheduler_type in benchmark_cpu_scheduler.BenchmarkSchedulerImplementations
-    ],
+    benchmark_cpu_scheduler.BenchmarkSchedulerImplementations.NO_SCHEDULING,
+    benchmark_cpu_scheduler.BenchmarkSchedulerImplementations,
     'The scheduler to use for choosing a core for running benchmarks.',
 )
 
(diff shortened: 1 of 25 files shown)


gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -141,3 +142,19 @@ def verify(self):
     cpu_mask = list(os.sched_getaffinity(0))
     if self._cpu_mask != cpu_mask:
       raise ValueError('Expected the CPU mask to not change.')
+
+
+class BenchmarkSchedulerImplementations(Enum):
+  NoScheduling = 1
+  Default = 2
```
ondrasej, line 149: Nit: In Python, enum value names should follow the naming for constants (`BIG_SNAKE_CASE`).

gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -141,3 +142,19 @@ def verify(self):
...
+
+def construct_benchmark_scheduler(
+    scheduler_type: BenchmarkSchedulerImplementations,
+) -> BenchmarkScheduler:
+  if scheduler_type == BenchmarkSchedulerImplementations.NoScheduling:
+    return NoSchedulingBenchmarkScheduler()
+  elif scheduler_type == BenchmarkSchedulerImplementations.Default:
+    return DefaultBenchmarkScheduler()
+  else:
+    raise ValueError('Unexpected Benchmark Scheduler Type.')
```
ondrasej, line 160: Optional: Since 3.10, there's a new `match` keyword that covers matching one expression against multiple values, e.g.:
```py
match scheduler_type:
  case BenchmarkSchedulerImplementations.NO_SCHEDULING:
    ...
  case BenchmarkSchedulerImplementtations.DEFAULT:
    ...
  case _:
    raise ValueError(...)
```

gematria/datasets/pipelines/benchmark_cpu_scheduler.py
```diff
@@ -17,6 +17,7 @@
 from collections.abc import Iterable
 import os
 import re
+from enum import Enum
```
ondrasej, line 20: The no-direct-imports rule applies here (only `collections.abc`, `typing`, and `typing_extensions` have an exception).

gematria/datasets/pipelines/benchmark_bbs.py
```diff
@@ -30,6 +31,15 @@
 _OUTPUT_FILE_PATTERN = flags.DEFINE_string(
     'output_file_pattern', None, 'The output file path/pattern.', required=True
 )
+_BENCHMARK_SCHEDULER = flags.DEFINE_enum(
```
ondrasej, line 34: Abseil has `flags.DEFINE_enum_class()` that allows defining flags from Python enums without any extra boilerplate.

ondrasej (changes requested):

ondrasej (approved):

PR #279, opened 2025-01-03
Port loss_utils_test to TF2

boomanaiden154:
The loss_utils class does not need any changes to support TF2, so this
patch simply migrates all the test cases to use eager execution. This
patch does not make the class idiomatically TF2 however. That will be a
later patch once everything else has been ported to allow for
incrementally improving the code while validating with all the tests.


gematria/model/python/loss_utils_test.py
```diff
@@ -216,173 +195,133 @@ def test_multi_task_with_mask(self):
...
+        1e-6,
+    )
+    self.assertAllEqual(percentiles, ((0, 0), (0.5, 0.5), (3, 4), (4, 6)))
 
   def test_single_task_unknown_shape(self):
     num_tasks = 1
-    actual_output = tf.placeholder(self.dtype, shape=(None, num_tasks))
-    expected_output = tf.placeholder(self.dtype, shape=(None, num_tasks))
+    actual_output = tf.reshape(self.actual_outputs_array, (-1, 1))
+    expected_output = tf.reshape(self.expected_outputs_array, (-1, 1))
```
ondrasej: As discussed offline - a reshape operation should not be needed here, the arrays should already have the right shape.

ondrasej (approved):

PR #280, opened 2025-01-04
Port model_base to TF2

boomanaiden154:
This patch ports most of the functionality of model_base to TF2 along
with all the tests. Notably left out is run_continuous_evaluation as it
is tested separately. This will be fixed in a future patch where the
tests for that are updated as well.


gematria/model/python/model_base.py
```diff
@@ -536,6 +440,44 @@ def output_tensor_names(self) -> Sequence[str]:
...
+        )
+
+      output = self._forward(feed_dict)
+
+      output_dict['output'] = tf.math.segment_sum(
+          output['output_deltas'],
+          feed_dict['delta_block_index'],
+          name=ModelBase.OUTPUT_TENSOR_NAME,
+      )
+      output_dict['output_deltas'] = output['output_deltas']
```
ondrasej, line 475: I think we can reuse `output` as the return value, just add `"output_deltas"` to it. This way, any other outputs returned by `_forward` will be preserved in the returned value (and the deltas + non-deltas will be more in line).

gematria/model/python/model_base.py
```diff
@@ -536,6 +440,44 @@ def output_tensor_names(self) -> Sequence[str]:
...
+
+      return output_dict
+    else:
+      return self._forward(feed_dict)
```
ondrasej, line 479: Optional: To reduce nesting:
```py
if self._use_deltas:
  return self._forward(feed_dict)

# No else needed.
```

gematria/model/python/model_base.py
```diff
@@ -1411,10 +1152,9 @@ def predict(
       with timer.scoped('ModelBase.predict - one batch'):
         schedule = self.schedule_batch(batch)
         if self._use_deltas:
-          output, output_deltas = sess.run(
-              (self._output_tensor, self._output_tensor_deltas),
-              feed_dict=schedule,
-          )
+          output_dict = self(schedule)
+          output = output_dict['output']
```
ondrasej, line 1156: These two lines are shared by the branches - I guess we can move them out and have just one `self(schedule)` statement for everything.

gematria/model/python/model_base.py
```diff
@@ -1531,74 +1268,120 @@ def run_one_epoch():
...
+        return self.train_batch(schedule)
 
-    with timer.scoped('ModelBase.train - one batch', num_iterations=num_epochs):
-      stats = None
-      while not monitored_session.should_stop():
-        stats = run_one_epoch()
-        logging.info('Training: %s', stats)
+    for _ in range(0, num_epochs):
+      stats = run_one_epoch()
+      logging.info('Training: %s', stats)
```
ondrasej, line 1275: Perhaps let's keep the timer here (unless it's replaced by something else later).

gematria/model/python/model_base.py
```diff
@@ -1531,74 +1268,120 @@ def run_one_epoch():
...
       return stats
```
ondrasej, line 1276: This should be indented by one less line (unless we keep the timer, which will fix the indentation another way).

gematria/model/python/model_base.py
```diff
@@ -1531,74 +1268,120 @@ def run_one_epoch():
...
 
+  def compute_loss(self, schedule: FeedDict):
```
ondrasej, line 1278: Please add a docstring (or mark this as private).

Also, please add a return type annotation.

gematria/model/python/model_base.py
```diff
@@ -1531,74 +1268,120 @@ def run_one_epoch():
...
+            self._loss_normalization, self._loss_type
+        )
+        loss_tensor = tf.reduce_mean(loss_tensor_per_task)
+
+        # The list of variables to optimize. By default, the list is empty which
+        # means optimize all trainable variables.
+        variables = set()
+        for variable_group in self._trained_variable_groups:
+          variables.update(
+              [
```
ondrasej, line 1341: Brackets shouldn't be really needed.

gematria/model/python/model_base.py
```diff
@@ -1531,74 +1268,120 @@ def run_one_epoch():
...
+        stats = {}
+        loss = self.compute_loss(schedule)
+        loss_tensor_per_task = loss.loss_tensor(
+            self._loss_normalization, self._loss_type
+        )
+        loss_tensor = tf.reduce_mean(loss_tensor_per_task)
+
+        # The list of variables to optimize. By default, the list is empty which
+        # means optimize all trainable variables.
+        variables = set()
```
ondrasej, line 1338: Let's use a different name for this set and the `variables` assigned below, the interaction between them may be too easy to overlook.

gematria/model/python/model_base.py
```diff
@@ -1531,74 +1268,120 @@ def run_one_epoch():
...
+      )
+
+      if self._grad_clip_norm:
+        if self._grad_clip_norm <= 0.0:
+          logging.warning(
+              'The gradients are clipped to zero. Please revise if this is not '
+              'intended.'
+          )
+        grads_and_vars = [
+            (self._clip_if_not_none(g), v) for g, v in grads_and_vars
```
ondrasej, line 1373: Eventually (not in this PR), we should check if we actually need to check for `None` here.

ondrasej (changes requested):

gematria/model/python/model_base.py
```diff
@@ -50,7 +50,7 @@
 # simpler than what is actually accepted by TensorFlow, but the typing should be
 # sufficient for our use. Moreover, since TensorFlow and NumPy do not provide
 # type annotations, both the key and the value are reduced to typing.Any.
-FeedDict = MutableMapping[tf.Tensor, np.ndarray]
+FeedDict = MutableMapping[str, Union[np.ndarray, tf.Tensor]]
```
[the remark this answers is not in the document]
boomanaiden154: Will do this systematically in a future PR.

gematria/model/python/model_base.py
boomanaiden154: Added to my TODO list.

gematria/model/python/model_base.py
```diff
@@ -536,6 +440,44 @@ def output_tensor_names(self) -> Sequence[str]:
...
+    outputs of the model given the inputs specified in feed_dict.
+    """
+    pass
+
+  def __call__(self, feed_dict, train=False):
+    """Implements the non-model specific part of the forward pass.
+
+    This function wraps the _forward method and does relevant calculations
+    when working with models that use deltas.
+    """
```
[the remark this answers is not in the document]
boomanaiden154: Added a new `OutputDict` instance of `TypedDict` to handle this. Inputs we can look at doing explicitly in the future.

PR #281, opened 2025-01-04
Port model_blocks to TF2

boomanaiden154:
This patch ports model_blocks and model_blocks_test to TF2. The major
change is the removal of the add_residual_connection_function. This is
not needed anymore given that everything will be constructed explicitly
eagerly inside the forward pass of models building off of the residual
connection layer.


gematria/model/python/model_blocks.py
```diff
@@ -79,36 +80,11 @@ def call(self, layer_inputs: tuple[tf.Tensor, tf.Tensor]) -> tf.Tensor:
     return tf.math.add(output_part, residual_part, name=self.name)
 
 
-def add_residual_connection(
-    output_part: tf.Tensor, residual_part: tf.Tensor, name: Optional[str] = None
-) -> tf.Tensor:
-  """Adds a residual connection to the output of a subnetwork.
+class CastLayer(tf_keras.layers.Layer):
```
ondrasej, line 83: Please add docstrings.

ondrasej (approved):

PR #284, opened 2025-01-04
Port inference to TF2

boomanaiden154:
This patch ports the inference library and associated unit tests to TF2.


gematria/model/python/inference_test.py
```diff
@@ -50,22 +50,11 @@ def __init__(self, use_custom_output_names=False, **kwargs):
     self.batch_sizes = []
     self.use_custom_output_names = use_custom_output_names
```
ondrasej: This (and the corresponding argument) can safely be removed, it's not used for anything.

ondrasej (approved):

PR #285, opened 2025-01-06
Port main_function to TF2

boomanaiden154:
This patch ports the main_function library to TF2. Quite a bit of code
had to be adjusted to use newer APIs that are not incompatible with TF2.
Some functionality was completely removed like exporting graph defs as
it makes significantly less sense with TF2 if it is even supported.


gematria/model/python/main_function.py
```diff
@@ -793,116 +732,111 @@ def run_gematria_model_from_command_line_flags(
...
-                input_graph_def=graph_def,
-                output_node_names=model.output_tensor_names,
-            )
-        tf.io.write_graph(
-            graph_def,
-            logdir=os.path.dirname(_GRAPH_DEF_FILE.value),
-            name=os.path.basename(_GRAPH_DEF_FILE.value),
+        blocks_with_throughput = _extract_basic_blocks_with_throughput(
+            model, basic_block_protos
         )
```
ondrasej, line 780: The `EXPORT_GRAPH_DEF` option was used to export models for TFLite conversion (this was the only way I was able to make work reliably with TF1).

We do not need to carry this over, but we need to keep a way to export models for TFLite consumption.

gematria/model/python/main_function.py
```diff
@@ -793,116 +732,111 @@ def run_gematria_model_from_command_line_flags(
...
+      )
+      tfrecord.write_protos(_GEMATRIA_OUTPUT_FILE.value, output_blocks)
+    elif _ACTION.value == model_options.Action.TRAIN:
+      if is_chief:
+        _resume_previous_experiment_if_needed()
+      if _WARMSTART_FILE.value:
+        # If there is a checkpoint to bootstrap from, we add an init_fn to the
+        # monitored session that restores it. This init_fn is called only when an
+        # actual checkpoint is not available to fully restore the model.
+        _warmstart_from_file(_WARMSTART_FILE.value)
```
ondrasej, line 813: I suppose this should be `_warmstart_fromm_file(model)`?

gematria/model/python/main_function.py
```diff
@@ -793,116 +732,111 @@ def run_gematria_model_from_command_line_flags(
...
+      elif _WARMSTART_DIR.value:
+        # If there is a directory to bootstrap from, we find the latest checkpoint
+        # in this directory and add an init_fn to the monitored session the same way
+        # as with _WARMSTART_FILE above.
+        _warmstart_from_dir(_WARMSTART_DIR.value)
```
ondrasej, line 818: I suppose this should be `_warmstart_from_dir(model)`?

gematria/model/python/main_function.py
```diff
@@ -793,116 +732,111 @@ def run_gematria_model_from_command_line_flags(
...
+        checkpoint_manager.save()
+
+      model.train(
+          tuple(blocks_with_throughput),
+          max_blocks_in_batch=_GEMATRIA_MAX_BLOCKS_IN_BATCH.value,
+          max_instructions_in_batch=max_instructions_in_batch,
+          num_epochs=_GEMATRIA_TRAINING_NUM_EPOCHS.value,
+          randomize_batches=_GEMATRIA_TRAINING_RANDOMIZE_BATCHES.value,
+          randomize_expected_outputs=randomize_expected_outputs,
+          hooks=[(_GEMATRIA_SAVE_CHECKPOINT_EPOCHS.value, checkpoint_model)],
```
ondrasej, line 841: Please add a TODO to write summaries during the training.

gematria/model/python/model_base.py
```diff
@@ -1210,6 +1210,7 @@ def train(
       max_instructions_in_batch: Optional[int],
       randomize_batches: bool = True,
       randomize_expected_outputs: bool = False,
+      hooks=[],
```
ondrasej, line 1213: Using a list (or any other mutable) object as a default arg value is potentially very risky. There's just a single instance of the mutable object created on parsing/compilation of the module that is then shared by all the calls (i.e. the same list object is used all the time).

The usual approach is to either take an immutable object (e.g. an empty tuple here) or just take `None` and then create the default object inside the function body (so that a new object is created for each call).

ondrasej, line 1213: Please add a type annotation.

gematria/model/python/model_base.py
```diff
@@ -1270,10 +1272,14 @@ def run_one_epoch():
         )
         return self.train_batch(schedule)
 
-    for _ in range(0, num_epochs):
+    for epoch_index in range(0, num_epochs):
       stats = run_one_epoch()
       logging.info('Training: %s', stats)
-      return stats
+      for hook in hooks:
+        epochs_every, hook_function = hook
```
ondrasej, line 1279: Nit:
```py
for epochs_every, hook_function in hooks:
  ...
```

gematria/model/python/model_base.py
```diff
@@ -1270,10 +1272,14 @@ def run_one_epoch():
...
+        if epoch_index % epochs_every == 0:
```
ondrasej, line 1280: We might want to something like `if (epoch_index + 1) % epoch_every == 0: ...` to avoid running the hooks after the first epoch.

gematria/model/python/main_function.py
boomanaiden154: Ack. We should have good support for saved models now, so I will put that on the list of things to do in the future.

gematria/model/python/main_function.py
```diff
@@ -295,15 +291,15 @@ def main(_):
...
+    100,
+    'The number of epochs of training after which a checkpoint is saved.',
 )
-_GEMATRIA_SAVE_SUMMARIES_SECS = flags.DEFINE_integer(
-    'gematria_save_summaries_secs',
-    60,
-    'The number of seconds of training after which summaries are saved.',
+_GEMATRIA_SAVE_SUMMARIES_EPOCHS = flags.DEFINE_integer(
+    'gematria_save_summaries_epochs',
+    100,
```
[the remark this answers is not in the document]
boomanaiden154: Added to the TODO list.

ondrasej (approved):

PR #288, opened 2025-01-19
Add support for storing basic block context in Gematria formats.

virajbshah:
* Add fields to the `proto` specification to store context.
 * Add members to the Gematria `BasicBlock` data structure to store context and update methods on it and its Python binding accordingly.
 * Bonus: Remove dangling TODO.

### DIFF
--- a/gematria/basic_block/basic_block.cc
+++ b/gematria/basic_block/basic_block.cc
@@ -377,11 +377,17 @@ std::ostream& operator<<(std::ostream& os, const Instruction& instruction) {
   return os;
 }
 
-BasicBlock::BasicBlock(std::vector<Instruction> instructions)
-    : instructions(std::move(instructions)) {}
+BasicBlock::BasicBlock(std::vector<Instruction> instructions,
+                       std::vector<Instruction> back_context,
+                       std::vector<Instruction> front_context)
+    : instructions(std::move(instructions)),
+      back_context(std::move(back_context)),
+      front_context(std::move(front_context)) {}
 
 bool BasicBlock::operator==(const BasicBlock& other) const {
-  return instructions == other.instructions;
+  return instructions == other.instructions &&
+         back_context == other.back_context &&
+         front_context == other.front_context;
 }
 
 std::string BasicBlock::ToString() const {
@@ -389,6 +395,24 @@ std::string BasicBlock::ToString() const {
   if (!instructions.empty()) {
     buffer += "instructions=InstructionList((";
     for (const Instruction& instruction : instructions) {
+      buffer += instruction.ToString();
+      buffer += ", ";
+    }
+    if (buffer.back() == ' ') buffer.pop_back();
+    buffer += "))";
+  }
+  if (!back_context.empty()) {
+    buffer += "back_context=InstructionList((";
+    for (const Instruction& instruction : back_context) {
+      buffer += instruction.ToString();
+      buffer += ", ";
+    }
+    if (buffer.back() == ' ') buffer.pop_back();
+    buffer += "))";
+  }
+  if (!front_context.empty()) {
+    buffer += "front_context=InstructionList((";
+    for (const Instruction& instruction : front_context) {
       buffer += instruction.ToString();
       buffer += ", ";
     }
--- a/gematria/proto/basic_block.proto
+++ b/gematria/proto/basic_block.proto
@@ -30,8 +30,26 @@ message BasicBlockProto {
   // same instruction.
   repeated CanonicalizedInstructionProto canonicalized_instructions = 2;
 
+  // An optional list of machine instructions preceding the basic block, used
+  // to provide context that lies before `canonicalized_instructions`. These
+  // instructions are not included in the timing measurements and predictions.
+  repeated CanonicalizedInstructionProto machine_back_context = 3;
+
+  // An optional list of machine instructions following the basic block, used
+  // to provide context lying after `canonicalized_instructions`. These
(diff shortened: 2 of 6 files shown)


virajbshah:
Bumping the C++ formatting CI version from `v4.11.0 (clang-format-16) -> v4.14.0 (clang-format-19)` should be enough to fix that failing check, though I'm not sure if re-running it after the update will use the new version without adding another commit to this PR.

boomanaiden154:
> though I'm not sure if re-running it after the update will use the new version without adding another commit to this PR.

#291. If you rerun the checks, it will be fixed. Github tests PRs as if they are merged against main by default.

gematria/proto/basic_block.proto
```diff
@@ -30,8 +30,26 @@ message BasicBlockProto {
...
+  // empty in case no back context is provided.
+  repeated CanonicalizedInstructionProto canonicalized_back_context = 5;
+
+  // Canonicalized instructions parallel to `machine_front_context`. May be
+  // empty in case no front context is provided.
+  repeated CanonicalizedInstructionProto canonicalized_front_context = 6;
+
   // The fingerprint-id of this basic block. Might be empty.
-  string fingerprint = 3;
+  string fingerprint = 7;
```
boomanaiden154, line 52: I don't think we have any major datasets of basic block protos written anywhere, but it would be nice to preserve backwards compatibility. Could you keep `fingerprint` where it was/assign it the same index?

gematria/basic_block/basic_block.cc
```diff
@@ -395,6 +401,24 @@ std::string BasicBlock::ToString() const {
     if (buffer.back() == ' ') buffer.pop_back();
     buffer += "))";
   }
+  if (!back_context.empty()) {
```
boomanaiden154, line 404: It doesn't look like there are any test changes for this functionality? Not sure there were any existing ones for `BasicBlock::ToString`, but it wouldn't be a bad idea to have some.

boomanaiden154 (commented):
Some initial comments.

virajbshah:
> If you rerun the checks, it will be fixed. Github tests PRs as if they are merged against main by default.

That would make sense, but for whatever reason re-running the action still uses `clang-format-18`. IIRC, adding another commit to the PR is what gets GitHub to pick up on the updated `workflows`.

boomanaiden154:
> That would make sense, but for whatever reason re-running the action still uses clang-format-18. IIRC, adding another commit to the PR is what gets GitHub to pick up on the updated workflows.

Interesting. Running `git merge main` on your branch after pulling a recent main should definitely fix the issue.

gematria/basic_block/basic_block.cc
ondrasej: +1, this should be tested at least a bit.

gematria/proto/basic_block.proto
ondrasej: +1, let's keep the field IDs stable as a standard practice.

gematria/proto/basic_block.proto
```diff
@@ -30,8 +30,26 @@ message BasicBlockProto {
   // same instruction.
   repeated CanonicalizedInstructionProto canonicalized_instructions = 2;
 
+  // An optional list of machine instructions preceding the basic block, used
+  // to provide context that lies before `canonicalized_instructions`. These
```
ondrasej, line 34: I'd s/lies before/precedes/.

The context doesn't have to be about code layout in the file, but this might be used just as well with execution traces, and the results would probably be even more interesting.

ondrasej, line 34: I'd s/lies before/precedes/.

The context doesn't have to be about code layout in the file, but this might be used just as well with execution traces, and the results would probably be even more interesting.

gematria/proto/basic_block.proto
```diff
@@ -30,8 +30,26 @@ message BasicBlockProto {
...
+  // instructions are not included in the timing measurements and predictions.
+  repeated MachineInstructionProto machine_back_context = 3;
+
+  // An optional list of machine instructions following the basic block, used
+  // to provide context lying after `canonicalized_instructions`. These
```
ondrasej, line 39: Similar to above, s/lying after/follows/.

ondrasej (changes requested):
Overall, this looks good. I'll approve once the comments are all resolved.

gematria/proto/basic_block.proto
```diff
@@ -30,8 +30,26 @@ message BasicBlockProto {
...
+  // instructions are not included in the timing measurements and predictions.
+  repeated MachineInstructionProto machine_front_context = 4;
+
+  // Canonicalized instructions parallel to `machine_back_context`. May be
+  // empty in case no back context is provided.
+  repeated CanonicalizedInstructionProto canonicalized_back_context = 5;
+
+  // Canonicalized instructions parallel to `machine_front_context`. May be
+  // empty in case no front context is provided.
+  repeated CanonicalizedInstructionProto canonicalized_front_context = 6;
```
ondrasej, line 49: Actually, let's rename these (and all the other related fields and variables) to `canonicalized_preceding_context` and `canonicalized_following_context`.

"front" and "back" might be interpreted both ways, so it's probably better to avoid them. I think I've seen cases where "front context" meant whatever precedes the main part (it's "in front of it") and "back context" to be what follows (similar to `std::vector`'s front and back).

ondrasej (changes requested):

ondrasej (approved):

boomanaiden154 (approved):

PR #289, opened 2025-01-19
Add support adding context instructions to basic block graphs.

virajbshah:
* Update the graph builder and its Python bindings to add context
   instructions to basic block graphs and store context node mask to later
   be used by models.
 * Add tests for the new graph builder functionality.

### DIFF
--- a/gematria/granite/graph_builder.cc
+++ b/gematria/granite/graph_builder.cc
@@ -190,7 +190,9 @@ BasicBlockGraphBuilder::BasicBlockGraphBuilder(
 }
 
 bool BasicBlockGraphBuilder::AddBasicBlockFromInstructions(
-    const std::vector<Instruction>& instructions) {
+    const std::vector<Instruction>& instructions,
+    const std::vector<Instruction>& back_context,
+    const std::vector<Instruction>& front_context) {
   if (instructions.empty()) return false;
   AddBasicBlockTransaction transaction(this);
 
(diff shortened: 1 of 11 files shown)


gematria/granite/graph_builder_test.cc
```diff
@@ -571,5 +575,97 @@ TEST_F(BasicBlockGraphBuilderTest, TwoNops) {
...
+      builder_->node_features(),
+      ElementsAre(TokenIndex("NOT"), TokenIndex("RCX"), TokenIndex("RCX"),
+                  TokenIndex("NOT"), TokenIndex("RCX"), TokenIndex("NOT"),
+                  TokenIndex("RCX"), TokenIndex("NOT"), TokenIndex("RCX"),
+                  TokenIndex("RCX"), TokenIndex("NOT"), TokenIndex("RCX")));
+  EXPECT_THAT(builder_->InstructionNodeMask(),
+              ElementsAre(true, false, false, true, false, true, false, true,
+                          false, false, true, false));
+  EXPECT_THAT(builder_->context_node_mask(),
+              ElementsAre(true, _, _, false, _, true, _, true, _, _, false, _));
```
ondrasej, line 645: I'd prefer to check all the values (to check that the context and non-context parts are contiguous).

ondrasej (changes requested):

gematria/granite/graph_builder.cc
```diff
@@ -202,58 +204,65 @@ bool BasicBlockGraphBuilder::AddBasicBlockFromInstructions(
...
 
-    // Add a structural dependency edge from the previous instruction.
-    if (previous_instruction_node >= 0) {
-      AddEdge(EdgeType::kStructuralDependency, previous_instruction_node,
-              instruction_node);
-    }
+      // Add a structural dependency edge from the previous instruction.
+      if (previous_instruction_node >= 0) {
+        AddEdge(EdgeType::kStructuralDependency, previous_instruction_node,
+                instruction_node);
```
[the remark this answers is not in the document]
ondrasej: We could give the new edge type(s) a more generic name, e.g. `kFromPrecedingContext` and `kToFollowingContext` so that it doesn't suggest any specific type of context. But what I'd prefer to do is to give the model a signal where the boundary is.

gematria/granite/graph_builder.cc
```diff
@@ -291,49 +302,53 @@ bool BasicBlockGraphBuilder::AddInputOperand(
   assert(instruction_node >= 0);
   assert(instruction_node < num_nodes());
 
+  bool is_context = context_node_mask_[instruction_node];
```
ondrasej, line 305: I'd rather pass this as an argument:
- it's probably faster (it will go through registers, with at least one fewer memory access),
- it reduces the risk that somewhere, sometimes we forget to set the the context mask for the instruction before adding its operands.

ondrasej:
> I have made some newer changes to the way `context_node_mask_` is populated. Previously, only nodes having `NodeType::kInstruction` were ensured to have their corresponding entries in `context_node_mask_` set correctly, as only these values mattered for usage in downstream models (and not checking the other values in `BasicBlockGraphBuilderTest::MultipleBasicBlocksWithContext` was to reflect this).
> 
> After this change, this behavior is better defined with `context_node_mask_` set to true for non-instruction nodes connected exclusively to context instruction nodes.
> 
> Please have a look.

I've added one technical comment, but otherwise this makes sense.

ondrasej (changes requested):

PR #290, opened 2025-01-19
Add support for using basic block context to GRANITE models.

virajbshah:
* Update seq2seq models to only use non-context nodes as input to the
   readout network.
 * Add tests for training seq2seq models with context.


boomanaiden154:
> LGTM overall. I'll approve once the parent ones are ready (merged).

This one is the lowest in the stack. It's the other ones that have dependencies.

PR #292, opened 2025-02-08
Add the "manual" tag to the access pattern microbenchmarks.

virajbshah:
* This should stop the microbenchmarks from being invoked by `bazel test ...`.
 * This prevents the `bazel build` CI checks from failing from one of the microbenchmarks timing out.
 * They can still be run individually using their fully qualified names.

### DIFF
--- a/gematria/experiments/access_pattern_bm/BUILD.bazel
+++ b/gematria/experiments/access_pattern_bm/BUILD.bazel
@@ -43,6 +43,7 @@ cc_test(
         "linked_list_test.cc",
     ],
     copts = BALANCE_FLUSHING_TIME_OPTS,
+    tags = ["manual"],
     deps = [
         ":linked_list",
         "@com_github_google_benchmark//:benchmark_main",
@@ -71,6 +72,7 @@ cc_test(
         "contiguous_matrix_test.cc",
     ],
     copts = BALANCE_FLUSHING_TIME_OPTS,
+    tags = ["manual"],
     deps = [
         ":contiguous_matrix",
         "@com_github_google_benchmark//:benchmark_main",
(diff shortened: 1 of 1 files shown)


boomanaiden154 (commented):
When I looked at this last, it seemed like there were options to set the maximum number of benchmark iterations. The benchmark not converging to a single value in CI seems to be what is actually causing the timeouts and setting the max number of iterations should limit the max time. Is it possible to do that rather than disabling all these tests in CI?

virajbshah:
IIUC, the number of benchmark iterations is governed by the time per iteration, such that the total CPU and wall times spent running all iterations of a benchmark cross some minimum threshold (0.5s of CPU time by default) regardless of whether the times per iteration converge or not.

This would suggest that a single iteration of `BM_STLAssocContainer_Flush<std::unordered_map<int, int>>/65536` (the benchmark within which we keep timing out) for some reason sometimes takes extraordinarily long to run on the CI runners, or alternatively, the first few iterations are much quicker than the later ones resulting in an unreasonably high estimated number of iterations.

That aside, I don't think there is a way to directly set a cap on the number of iterations. We can, however, have all variants of a given benchmark run for a fixed number of iterations or set a lower time threshold which would indirectly reduce, but not limit, the number of benchmark iterations.

Another option would be configure Bazel to pass `--benchmark_dry_run` whenever it runs under CI, which would have it run one iteration of each benchmark so that we know they don't fail, otherwise we'd have to limit the benchmark size to the one below `.../65536` if we want to keep running it under CI.

boomanaiden154:
> Another option would be configure Bazel to pass --benchmark_dry_run whenever it runs under CI, which would have it run one iteration of each benchmark so that we know they don't fail, otherwise we'd have to limit the benchmark size to the one below .../65536 if we want to keep running it under CI.

Ah, didn't realize that was an option. I think this would be the ideal solution. Getting actual benchmark numbers is meaningless on CI, but this ensures they still work when we do want to execute them.

ondrasej:
SGTM for running just one iteration of the benchmark under CI. There's still some value in running through the code even if we don't collect any data.

virajbshah:
Dropped in favor of #294 .

PR #294, opened 2025-02-10
Configure microbenchmarks to dry-run under CI.

virajbshah:
* Adds a `--define run_under_ci=1` to the `bazel test` CI invocation.
 * Benchmarks are dry-run when `--define run_under_ci=1` is set.


boomanaiden154 (approved):
This seems like a reasonable enough approach to me, but please wait for @ondrasej to review as I'm not ultra familiar with the best way to do things in bazel.

virajbshah:
I realized that the benchmarks should also be able to take arguments from the environment, i.e. we could `bazel test --action_env=BENCHMARK_DRY_RUN=1 ...` without requiring any special handling in `BUILD.bazel`. The downside being that using environment variables this way seems like a generally problematic/hard-to-debug way of doing things, and would apply to every test (which shouldn't currently be a real issue because nothing else is listening for `BENCHMARK_DRY_RUN`, but still). Any thoughts?

boomanaiden154:
I think explicitly handling it is better than passing `--action_env` for the reasons you mentioned.

ondrasej (approved):

PR #296, opened 2025-03-11
Add TensorFlow profiler to training loop.

virajbshah:
Adds some flags allowing the training loop to be run under the TensorFlow profiler. The profiler is part of the TF 2.x API but should (and appears to) work with TF 1.x models in this way.

### DIFF
--- a/gematria/model/python/main_function_test.py
+++ b/gematria/model/python/main_function_test.py
@@ -17,7 +17,7 @@ import functools
 from os import path
 import os
 import re
-from threading import Thread
+import threading
 from unittest import mock
 
 from absl import flags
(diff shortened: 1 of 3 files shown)


gematria/model/python/main_function_test.py
```diff
@@ -831,6 +833,72 @@ def test_multi_task_flags(self):
...
+    """
+    num_epochs = 10
+    max_blocks_in_batch = 15
+    max_instructions_in_batch = 124
+    learning_rate = 0.321
+    randomize_batches = False
+    training_throughput_selection = io_options.ThroughputSelection.RANDOM
+    checkpoint_dir = path.join(self.work_directory.full_path, 'checkpoint')
+    summary_dir = path.join(self.work_directory.full_path, 'summary')
+    tf_profiler_port = 6009
```
ondrasej, line 852: Optional: If this works reliably in CI, this is fine. Otherwise, we could use some randomization to find an unused port for the test.

ondrasej (approved):
Thanks for adding this!

PR #297, opened 2025-03-19
Fix use after free in annotating importer

boomanaiden154:
There is a line in annotating importer that attempts to get the base name of a path from a StringRef. Before this patch, a StringRef was referenced, .str() was called on it, and then it was passed into a function to extract the base name that took in a StringRef and returned a StringRef. Calling .str() caused a temporary string to get created. Given that string is a temporary, it can get deallocated after calling the function to get the basename, resulting in a use after free when the StringRef is referenced later. This patch simply removes the .str() call so that we purely operate with StringRefs to proper backing memory, which fixes the issue.

This fixes the Gematria->Google3 import. Not entirely sure how/why this did not fail in the open.


ondrasej (approved):
More efficient and more correct at the same time :)

PR #298, opened 2025-03-22
Bump LLVM

boomanaiden154:
This patch bumps LLVM to the latest version as of 3/22/25. I wanted to see if there were any API updates/changes impacting us, but doesn't seem like there are. Given that, might as well push the update.


virajbshah:
We should also update `LLVM_COMMIT` in the CI config to match this.

boomanaiden154:
> We should also update LLVM_COMMIT in the CI config to match this.

Good point. Fixed.

ondrasej (approved):

PR #301, opened 2025-04-01
Make Exegesis Benchmark Retry Benchmarking Upon Perf Counter Failure

boomanaiden154:
This patch makes exegesis_benchmark_lib retry benchmarking if the perf
counter is not enabled the entire time. This is part of the fix for zero
performance counter values on internal machines. Currently this
reassembles the snippet everytime a failure occurs, which could incur a
performance penalty. Given the interference patterns we have seen so far
though, this should not be a major issue. If this part of the code ends
up becoming a bottleneck, we can readjust later. The ideal place to do
this performance-wise is inside the BenchmarkRunner implementation
itself, but I want to avoid adding complexity there if we can avoid it.

### DIFF
--- a/gematria/datasets/exegesis_benchmark_lib.cc
+++ b/gematria/datasets/exegesis_benchmark_lib.cc
@@ -51,6 +51,13 @@ using namespace llvm;
 using namespace llvm::exegesis;
 
 namespace gematria {
+namespace {
+
+// The maximum number of attempts to benchmark a snippet retrying when
+// performance counters are only partially enabled before erroring out.
+constexpr unsigned kMaxBenchmarkingAttempts = 1000;
+
+}  // namespace
 
 ExegesisBenchmark::ExegesisBenchmark(LLVMState &&State)
     : ExegesisState(std::move(State)) {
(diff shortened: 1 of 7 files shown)


gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -340,35 +341,43 @@ Expected<BenchmarkCode> ExegesisBenchmark::processAnnotatedBlock(
...
+    unsigned int MinInstructions, std::optional<int> BenchmarkProcessCPU) {
+  while (true) {
+    Expected<BenchmarkRunner::RunnableConfiguration> RC =
+        BenchRunner->getRunnableConfiguration(BenchCode, 5000, 0, Repetitor);
+    if (!RC) return RC.takeError();
+
+    std::pair<Error, Benchmark> BenchmarkResultOrErr =
+        BenchRunner->runConfiguration(std::move(*RC), {}, BenchmarkProcessCPU);
+    if (std::get<0>(BenchmarkResultOrErr).isA<PerfCounterNotFullyEnabled>())
+      continue;
```
ondrasej, line 355: We should probably have a limit on the number of retries (perhaps make this a parameter?), to avoid looping forever because of some weird corner case in the counter implementation.

gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -340,35 +341,43 @@ Expected<BenchmarkCode> ExegesisBenchmark::processAnnotatedBlock(
   return BenchmarkConfiguration;
 }
 
+Expected<Benchmark> ExegesisBenchmark::benchmarkConfiguration(
```
ondrasej, line 344: Nit: The function name looks as if it returns a benchmark configuration (without running anything). Maybe we should rename this to `runBenchmarkForConfiguration()` or something similar?

gematria/datasets/exegesis_benchmark_lib.cc
```diff
@@ -340,35 +341,43 @@ Expected<BenchmarkCode> ExegesisBenchmark::processAnnotatedBlock(
...
+    const BenchmarkCode &BenchCode, const SnippetRepetitor &Repetitor,
+    unsigned int MinInstructions, std::optional<int> BenchmarkProcessCPU) {
+  while (true) {
+    Expected<BenchmarkRunner::RunnableConfiguration> RC =
+        BenchRunner->getRunnableConfiguration(BenchCode, 5000, 0, Repetitor);
```
ondrasej, line 349: s/5000/MinInstructions/

ondrasej (changes requested):

gematria/datasets/exegesis_benchmark_lib.cc
boomanaiden154: Thanks for the catch. That would've been frustrating to deal with if it didn't get caught before running the pipeline...

ondrasej (approved):

PR #313, opened 2025-04-14
Add `cosine_decay_restarts` to available LR schedules.

virajbshah:
Adds the `tf.train.cosine_decay_restarts` learning rate schedule.


ondrasej (approved):
Nice!

PR #314, opened 2025-04-14
Add histogram summaries to `GnnModelBase`.

virajbshah:
* Adds a call to a new abstract method to set up histogram summaries to the model initialization pipeline.
 * Implements this in `GnnModelBase` to log histogram summaries for all of its trainable variables, i.e. those in the graph network layers.

### DIFF
--- a/gematria/granite/python/gnn_model_base.py
+++ b/gematria/granite/python/gnn_model_base.py
@@ -417,6 +417,15 @@ class GnnModelBase(model_base.ModelBase):
           )
     return graphs_tuple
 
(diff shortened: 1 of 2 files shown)


boomanaiden154 (commented):
What exactly is the purpose of logging the weight distributions like this?

virajbshah:
It's supposed to be an extra diagnostic to catch things like some weights exploding or gradients vanishing. I added it to compare embedding distributions between regular and context nodes.

boomanaiden154 (approved):
LGTM, but it's probably good to wait for Ondrej's approval as well.

PR #315, opened 2025-04-14
Switch all uses of pybind's `py::enum_` to `py::native_enum`.

virajbshah:
* Switches the following uses of the now deprecated `py::enum_` to `py::native_enum`:
   * `basic_block.OperandType`
   * `graph_builder.NodeType`
   * `graph_builder.EdgeType`
   * `oov_token_behavior.BehaviorType`
   * `bhive_to_exegesis.AnnotatorType`
 * Updates any usages of these types to take advantage of the full native `Enum` interface.
 * The downside is that `py::native_enum` does not take an `Enum`-level docstring. Member-level docstrings are kept.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -141,9 +141,9 @@ git_repository(
 
 new_git_repository(
     name = "pybind11",
+    commit = "e7e5d6e5bb0af543a2ded6d34163176c3e6ab745",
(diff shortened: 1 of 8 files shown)


virajbshah:
It seems like bumping `pybind11` causes some incompatibility with TensorFlow's usage of `pybind11_absl`. It'll probably take some investigating to figure out.

PR #316, opened 2025-04-14
Remove the hack used while importing third party python libraries.

virajbshah:
* The `__init__.py` hack no longer seems to be necessary to import third party python libraries by name: Bazel appends the appropriate paths to `sys.path` automatically based on the package deps.
 * In fact, `__init__.py` is not used at all under Bazel or within Bazel built wheels as it is not listed in any deps and hence does not appear under any module's runfiles.


ondrasej (approved):
Cool! I'm happy this is no longer needed.

PR #326, opened 2025-04-21
Restore TensorBoard summary logging after TF 2 migration.

virajbshah:
* Logs most of the previously logged scalars.
 * Bonus: wraps training in `timer.scoped`.


virajbshah:
If preferred, I can add delta loss summaries to this PR, otherwise left as a TODO. This might also be a good opportunity to reconsider whether or not delta losses should be logged for `seq2seq` models not using the `seq2seq_loss`.

boomanaiden154 (approved):
I figured it was something like this. Thanks for the patch!

PR #329, opened 2025-04-21
Remove tf.compat.v1 imports in io/

boomanaiden154:
The only thing that we needed to keep was the usage of
tf.compat.v1.io.tf_record_iterator that should get removed eventually
when we move over to the actual dataset APIs.

### DIFF
--- a/gematria/granite/python/BUILD.bazel
+++ b/gematria/granite/python/BUILD.bazel
@@ -89,6 +89,7 @@ gematria_py_test(
     size = "small",
     timeout = "moderate",
     srcs = ["graph_builder_model_base_test.py"],
+    shard_count = 25,
     deps = [
         ":gnn_model_base",
         ":graph_builder",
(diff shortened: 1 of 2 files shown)


gematria/io/python/tfrecord.py
```diff
@@ -49,7 +49,7 @@ def read_protos(
     # and do what the user expects.
     filenames = (filenames,)
   for filename in filenames:
-    for raw_record in tf.io.tf_record_iterator(filename):
+    for raw_record in tf.compat.v1.io.tf_record_iterator(filename):
       yield proto_class.FromString(raw_record)
```
ondrasej, line 53: Optional: We could also get rid of the `compat.v1` reference here in the code with something like:
```py
yield from tf.data.TFRecordDataset(filenames).map(proto_class.FromString)
```
instead of the two nested for loops.

ondrasej (approved):

gematria/io/python/tfrecord.py
boomanaiden154: I think I'll leave this to a future patch because I'm planning on rewriting most of the dataset stuff in the near future anyways.

PR #333, opened 2025-04-22
Misc fixes for google import

boomanaiden154:
This patch contains some misc fixes most related to type checking that enable us to import gematria into google after the TF2 migration. Most are pretty standard. We disable an attribute error on run_continuous_evaluation mostly to silence it for now as tf_slim is not really compatible with TF2 and needs to be rewritten.

### DIFF
--- a/gematria/model/python/model_base.py
+++ b/gematria/model/python/model_base.py
@@ -1106,7 +1106,7 @@ class ModelBase(tf.Module, metaclass=abc.ABCMeta):
         tf_slim.evaluation.SummaryAtEndHook(summary_dir, feed_dict=schedule),
         # Save the models with the best MAPE.
         SaveBestCheckpoint(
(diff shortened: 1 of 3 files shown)


gematria/model/python/model_base.py
```diff
@@ -1106,7 +1106,7 @@ def run_continuous_evaluation(
         tf_slim.evaluation.SummaryAtEndHook(summary_dir, feed_dict=schedule),
         # Save the models with the best MAPE.
         SaveBestCheckpoint(
-            error_tensor=self._loss.mean_absolute_percentage_error,
+            error_tensor=self._loss.mean_absolute_percentage_error,  # pytype: disable=attribute-error
```
ondrasej, line 1109: Please add a short comment explaining what's wrong here.

ondrasej (approved):

PR #339, opened 2025-04-28
Add option to compile_modules pipeline to ingest raw BBs

boomanaiden154:
This patch adds an option to the compile_modules pipeline to ingest
basic blocks in a hex file (like in the BHive dataset) rather than
compiling IR stored in parquet. This enables us to ingest a slightly
processed CSV from the BHive dataset to then annotate and benchmark it
with our pipeline to ensure the numbers match up reasonably well.

### DIFF
--- a/gematria/datasets/pipelines/compile_modules_lib.py
+++ b/gematria/datasets/pipelines/compile_modules_lib.py
@@ -231,13 +231,14 @@ class GetVocab(beam.DoFn):
 
 
 def get_bbs(
-    input_file_pattern: str,
+    input_file_pattern: str | None,
     output_file: str,
     remove_memory_accessing_instructions: bool,
     annotator_type: bhive_to_exegesis.AnnotatorType,
     max_annotation_attempts: int,
     vocab_output_file: str,
     skip_no_loop_register: bool,
+    input_hex_bbs_file_pattern: str | None,
 ) -> Callable[[beam.Pipeline], None]:
   """Creates a pipeline to process BBs from IR modules.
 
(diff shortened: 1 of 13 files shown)


gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -259,34 +261,52 @@ def get_bbs(
...
     )
-    lowered_modules = optimized_modules | 'Lower' >> beam.ParDo(
-        LowerModulesAsm(['-O0', '-O1', '-O2', '-O3'])
-    )
-    bb_hex_values = lowered_modules | 'Get BBs' >> beam.ParDo(
-        GetBBsFromModule()
-    )
-    bb_hex_values_deduplicated = (
-        bb_hex_values | 'Deduplicate' >> DeduplicateValues()
+  if input_file_pattern and input_hex_bbs_file_pattern:
```
ondrasej, line 277: I'd go for an explicit `is not None` here, to be symmetrical with the check above.

Without it, there are cases like when both args contain an empty string that would pass both checks and then cause confusion below. When `None` is concerned, the Google style guide recommends using the explicit `is [not] None`, so I'd go with that rather than the implicit conversions.

Btw. the simplest way to check this is to compare the boolean values obtained form both args:
```py
if (input_filt_pattern is None) == (input_hex_bbs_file_pattern is None):
  raise ValueError('Exactly one of input_file_pattern and input_hex_bbs_file_pattern must be set')
```
This way, the exception is raised when both are `None` or both are `not None`.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -259,34 +261,52 @@ def get_bbs(
...
+    raise ValueError(
+        'Only one of input_file_pattern or input_hex_bbs_file_pattern can be'
+        ' set.'
     )
+
+  def pipeline(root: beam.Pipeline) -> None:
+    if input_hex_bbs_file_pattern:
```
ondrasej, line 284: `input_hex_bbs_file_pattern is not None`, following the same argument as above.

Same for the other checks below.

gematria/datasets/pipelines/compile_modules_lib.py
```diff
@@ -259,34 +261,52 @@ def get_bbs(
...
+      bb_hex_values_deduplicated = root | 'Read' >> beam.io.ReadFromText(
+          input_hex_bbs_file_pattern
+      )
```
ondrasej, line 287: I'd also add a `DeduplicateValues()` step here, just in case. We say de-duplicated, but there's no guarantee that the input file doesn't contain duplicates.

gematria/datasets/pipelines/compile_modules_lib_test.py
```diff
@@ -198,6 +198,45 @@ def test_get_bbs(self, annotator_type):
...
+
+    pipeline_constructor = compile_modules_lib.get_bbs(
+        None,
+        output_file_pattern,
+        False,
+        bhive_to_exegesis.AnnotatorType.exegesis,
+        50,
+        vocab_output_file_pattern,
+        False,
+        test_bb_file.full_path,
```
ondrasej, line 239: At this argument count, it's probably worth using keyword args for all of them.

ondrasej (changes requested):

gematria/datasets/pipelines/compile_modules_lib_test.py
boomanaiden154: Updated this callsite and the others that had to be adjusted due to the new argument.

ondrasej (approved):

PR #342, opened 2025-05-17
Remove model_ref in token_graph_builder_model

boomanaiden154:
With the port to TF2 I added a hack to get access to instruction
annotations and the node mask within the call function of
TokenGraphBuilderModelNodeEmbed. This was necessary due to TF2
preferring eager mode so we could not just pass tensor references
around.

This patch makes everything more canonically TF2 by moving the data
around through standard dataflow (with some slight complexity to ensure
things are getting passed around to the right places) rather than
passing references around. Either one should theoretically work with
tf.function annotations depending upon scope, but this will definitely
work.

### DIFF
--- a/gematria/granite/python/token_graph_builder_model.py
+++ b/gematria/granite/python/token_graph_builder_model.py
@@ -345,7 +345,6 @@ class TokenGraphBuilderModel(graph_builder_model_base.GraphBuilderModelBase):
                     vocab_size=len(self._token_list),
                     common_embed_dim=self._common_node_embedding_size,
                     num_annotations=self._num_annotations,
-                    model_ref=self,
                     initializer=embedding_initializer,
                 ),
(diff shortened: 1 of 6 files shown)


gematria/granite/python/token_graph_builder_model.py
```diff
@@ -364,6 +363,10 @@ def _create_graph_network_modules(
             num_iterations=1,
             layer_normalization=options.EnableFeature.NEVER,
             residual_connection=options.EnableFeature.NEVER,
+            extra_node_inputs=[
```
ondrasej, line 366: Nit: Make this a tuple rather than a list (to make it immutable).

ondrasej (approved):
Cool!

I think we could simplify the models quite a bit by removing the layers abstraction, and writing this directly in the model. But that is for some other time.

PR #343, opened 2025-05-18
Refactor loss_utils to be idiomatically TF2

boomanaiden154:
This patch refactors loss_utils to be more TF2 like. The main impact of
this is that LossComputation now inherits from
tf.experimental.ExtensionType which implies everything about the class
has to be setup in the constructor. This enables returning a
LossComputation object from a function decorated with tf.function, which
enables us to wrap the entire training step in tf.function.


gematria/model/python/model_base.py
```diff
@@ -1308,28 +1307,28 @@ def _compute_loss(self, schedule: FeedDict) -> loss_utils.LossComputation:
...
           dtype=self.dtype,
+          normalization=self._loss_normalization,
+          loss_type=self._loss_type,
       )
 
       if self._use_delta_loss:
-        loss = self._delta_loss
+        loss = delta_loss
 
     return loss
```
ondrasej, line 1328: Now that we do not store `loss` and `delta_loss` as attributes, there's no need to create both objects, and the whole function can be simplified to:
```py
def _compute_loss(self, schedule: FeedDict) -> loss_utils.LossComputation:
  output = self(schedule, train=True)
  maybe_deltas = '_deltas' if self._use_deltas else ''
  return loss_utils.LossComputation(
      output[f'output{maybe_deltas}'],
      tf.constant(schedule[f'expected_outputs{maybe_deltas}']),
      tf.constant(schedule[f'output_mask{maybe_deltas}']),
      percentile_ranks=self._collected_percentile_ranks,
      dtype=self.dtype,
      normalization=self._loss_normalization,
      loss_type=self._loss_type,
  )
```

I suppose that in JIT mode, all the unused objects would be optimized out, but if we can get less code, that's even better.

gematria/model/python/model_base.py
```diff
@@ -335,7 +335,6 @@ def __init__(
 
     self._decayed_learning_rate = None
     self._loss: Optional[loss_utils.LossComputation] = None
-    self._delta_loss: Optional[loss_utils.LossComputation] = None
     self._train_step: Optional[tf.Operation] = None
     self._optimizer: Union[
         tf.train.Optimizer, tf.train.SyncReplicasOptimizer
```
ondrasej, line 340: Unrelated to this PR, but if you feel like it:
```py
self._loss: loss_utils.LossComputation | None = None
self._train_step: tf.Operation | None = None
self._optimizer: tf.train.Optimizer | tf.tran.SyncRepliccasOptimizer | None = None
```

gematria/model/python/loss_utils.py
```diff
@@ -72,225 +77,250 @@ def __init__(
...
 
-  def _make_percentile_tensor(self, values: tf.RaggedTensor) -> tf.Tensor:
+  def _make_percentile_tensor(
+      self,
+      values: tf.RaggedTensor,
+      num_tasks: int,
+      percentile_ranks: Sequence[int],
+      dtype: tf.dtypes.DType,
+  ) -> tf.Tensor:
     """Creates a percentile tensor from 'values' using self.percentile_ranks.
```
ondrasej, line 254: Remove `self.` from the line.

gematria/model/python/loss_utils.py
```diff
@@ -72,225 +77,250 @@ def __init__(
...
-    """Returns the percentiles of the absolute error."""
-    return self._absolute_error_percentiles
-
-  @property
-  def absolute_percentage_error_percentiles(self) -> tf.Tensor:
-    """Returns the percentiles of the absolute percentage error."""
-    return self._absolute_percentage_error_percentiles
-
-  def loss_tensor(
+  def _loss_tensor(
```
ondrasej, line 171: Here and below: this doesn't need to be a method - `self` is used only to call other "methods" of the class, but never for accessing any attributes of the class, that is done only in `__init__`.

I think it would make sense to turn all of these into regular functions (maybe private, maybe public) and make `LossComputation` as small as possible. That should also allow reusing them in other contexts and testing them.

gematria/model/python/loss_utils.py
```diff
@@ -72,225 +77,250 @@ def __init__(
...
       normalization: options.ErrorNormalization,
       loss_type: options.LossType,
+      num_tasks: int,
+      dtype: tf.dtypes.DType,
+      delta: tf.Tensor,
+      squared_errors: tf.Tensor,
+      absolute_errors: tf.Tensor,
+      absolute_percentage_errors: tf.Tensor,
+      squared_percentage_error: tf.Tensor,
+      absolute_expected_outputs_or_one: tf.Tensor,
```
ondrasej, line 182: Please add all of these to the docstring.

gematria/model/python/loss_utils.py
```diff
@@ -72,225 +77,250 @@ def __init__(
...
+        return absolute_errors
       case options.ErrorNormalization.PERCENTAGE_ERROR:
-        return self._absolute_percentage_errors
+        return absolute_percentage_errors
       case options.ErrorNormalization.EXPECTED_VALUE_GREATER_THAN_ONE:
-        return self._absolute_errors / self._absolute_expected_outputs_or_one
+        return absolute_errors / absolute_expected_outputs_or_one
       case _:
         raise NotImplementedError(
             'Absolute errors not implemented yet: {normalization}'
```
ondrasej, line 326: These functions are now almost the same (the only difference is in the `tf.square()` call + minor style differences). I think we could move the `tf.square()` call up the stack, keep just one version, and rename it to `_apply_normalization()`.

gematria/model/python/loss_utils.py
```diff
@@ -20,21 +20,26 @@
...
 
+  loss_tensor: tf.Tensor
+  mean_absolute_error: tf.Tensor
+  mean_squared_error: tf.Tensor
+  mean_absolute_percentage_error: tf.Tensor
+  mean_squared_percentage_error: tf.Tensor
+  absolute_error_percentiles: tf.Tensor
+  absolute_percentage_error_percentiles: tf.Tensor
+
   def __init__(
```
ondrasej, line 34: IIUC, `ExtensionType.__init__` is not really meant to be overridden (similar to dataclasses, it does some init auto-generation magic).

IMO a safer and more way is to turn this into a factory `@classmethod`, something like
```py
@classmethod
def create(
    output_values: tf.Tensor,
    expected_outputs: tf.Tensor,
    ...,
) -> Self:
  ...
```

This would go very well with the proposed change of methods to module-level functions.

ondrasej (approved):
I think we can go a bit further and simplify the code a bit more. A lot of things that happened in the old version were there to share sub-results across the computation. Since we no longer do that (and perhaps depend on the JIT to optimize them away?), we can go a bit further and make the code quite a bit shorter.

ondrasej (changes requested):
Clicked on "Approve" too fast.

gematria/model/python/model_base.py
boomanaiden154: It's on my TODO list to do a sweeping change that gets all of these, so I'd prefer to leave it for that.

boomanaiden154:
> I think this could still be simplified quite a bit: _loss_tensor and _apply_normalization do very few tensor ops, otherwise they just select from their args based on other arg values (lot of the logic that is there was to make things work in the TF1 graph building world).

Yeah, not super satisfied with this. Letting us use `tf.function` is a big improvement though.

> But maybe let's merge this and I'll take a stab at the simplifications once the current version is merged.

Sounds good.

PR #346, opened 2025-05-19
Migrate to Bazel 7.

virajbshah:
* Migrate to Bazel 7, but continue using `WORKSPACE`.
 * Bump a number of packages to make everything Bazel 7 compatible.
 * Begin using a hermetic Python3.11 toolchain.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -142,4 +142,7 @@ jobs:
     - name: Run build
       run: bazel --output_base=/mnt/output_base build -c ${{ matrix.build-config }} ...
     - name: Run tests
-      run: bazel --output_base=/mnt/output_base test -c ${{ matrix.build-config }} --define run_under_ci=1 --test_tag_filters="-perf_counters" --test_output=errors ...
+      # TODO(vbshah): We exclude `requirements_3_X_test` because its timeout is
+      # too short for it to pass consistently. Re-include it once it is fixed
+      # or `rules_python` is patched to extend the timeout. 
+      run: bazel --output_base=/mnt/output_base test -c ${{ matrix.build-config }} --define run_under_ci=1 --test_tag_filters="-perf_counters" --test_output=errors -- ... -//:requirements_3_12_test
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.11'
```
boomanaiden154, line 40: Is there a reason to not go straight to `3.12`? All the code works on 3.12 (minus maybe some external dependency versions) given that's what is used in google3.

requirements_lock_3_11.txt
```diff
@@ -0,0 +1,1560 @@
+#
+# This file is autogenerated by pip-compile with Python 3.11
+# by the following command:
+#
+#    bazel run //:requirements_3_11.update
+#
+absl-py==2.2.2 \
```
boomanaiden154, line 7: Does this need to be made version specific? Most of the time a lockfile generated for an older python version will work with newer versions as well. Although I'm not sure how that interacts with the hermetic python toolchain.

.github/workflows/main.yaml
virajbshah: Really, the motivation for supporting 3.11 (or even 3.10) is that registering multiple Python toolchains should be fairly straightforward (replace `python_register_toolchains` with `python_register_multi_toolchains` and providing a list of supported versions).

I believe there isn't any significant advantage to doing so unless some other project requiring 3.11 pulls this in as a Bazel dependency. Another small potential advantage would being able to set up newer toolchains before they are internally supported while 3.12 remains the default.

ondrasej: Some Python packages are a bit picky about Python versions (I'm looking at you, TensorFlow). So, to keep dependencies simple, I'd stick with just one version (which would be 3.12, to be in line with the internal version).

ondrasej (approved):

PR #347, opened 2025-05-20
Bump LLVM to 88c4ef2f9fc0cda90c8452bc1c46844aaa722a3e

boomanaiden154:
This includes https://github.com/llvm/llvm-project/pull/137581 which makes it a bit nicer to deal with verification failures, some of which come from inline assembly and would be difficult to fix within LLVM.


orodley (approved):

boomanaiden154:
Tested locally and everything passes.

PR #348, opened 2025-05-21
[CI] Move the Bazel output base to a free disk.

virajbshah:
* The CI for #346 fails because the runner runs out of disk space.
 * CI runners should have a disk with plenty of free capacity mounted on `/mnt`.
 * This PR moves the Bazel output base there, which should keep us from running into this issue again.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -125,11 +125,13 @@ jobs:
         sudo rm -f /usr/bin/clang++ && \
         sudo ln -s /usr/bin/clang-19 /usr/bin/clang && \
         sudo ln -s /usr/bin/clang-19 /usr/bin/clang++
+    - name: Setup bazel output base
+      run: sudo mkdir /mnt/output_base && sudo chmod a+rw /mnt/output_base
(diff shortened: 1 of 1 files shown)


boomanaiden154 (commented):
Are we brushing up against the threshold already?

And why do we increase disk space usage? The new hermetic python toolchain?

virajbshah:
From my local Bazel cache, it looks like:
 * The hermetic python toolchain uses up some extra space because of the runfiles mechanism creating a bunch of symlinks for each PyPI package, which is negligible for most packages, but adds up to ~75M (per `gematria_py_binary` that uses it) for TensorFlow. I'm guessing this adds up to a couple gigabytes.
 * The `pip install -r ...` is also still kept in (though should no longer be necessary), so we have two copies of every package (one from this, and one Bazel-managed copy).
 * I'm not sure if this is new behavior from after the `rules_python` or `pybind11_bazel` bumps, but Bazel also copies over every pybind11 extension module `*.so` into a `*.pyd`.
 * The repo cache from the new `http_archive` rules is also pretty big (~6G).

PR #350, opened 2025-05-29
Simplified the parameterized loss computation.

ondrasej:
Streamlines the computation of the loss and brings it more in line with an eager
TF2 style. 

As a side effect, I've added some duplication of work in repeated normalization of
the outputs. This could be avoided by unwrapping some of the computation, but it's
probably not worth the extra code and complexity. And, at least in theory, the
duplicate ops should be removed by the JIT graph optimizer.

This is a follow-up to ##343.

### DIFF
--- a/gematria/model/python/loss_utils.py
+++ b/gematria/model/python/loss_utils.py
@@ -87,7 +87,7 @@ def _apply_loss_function(
       return tf.abs(normalized_delta)
     case options.LossType.HUBER:
       return _huber(tf.abs(normalized_delta))
-  raise ValueError('Unexpected loss type {loss_type!r}')
+  raise ValueError(f'Unexpected loss type {loss_type!r}')
 
 
 def _apply_normalization(
@@ -99,7 +99,6 @@ def _apply_normalization(
 
   Args:
     normalization: The selected normalization function.
-    num_tasks: The number of tasks used in the model.
     delta: The deltas between the actual output and the expected output.
     expected_outputs: The expected output values.
 
@@ -127,7 +126,7 @@ def _loss_tensor(
     expected_outputs: tf.Tensor,
 ) -> tf.Tensor:
   """Creates a loss tensor for the given loss type.
-  
+
   Args:
     normalization: The error normalization used in the loss tensor.
     loss_type: The loss function used in the loss tensor.
@@ -136,7 +135,7 @@ def _loss_tensor(
       expected outputs. Must have have shape (None, num_tasks).
     expected_outputs: A tensor that contains the expected outputs. Must have
       shape (None, num_tasks).
-  
+
   Returns:
     A tensor that contains the loss computed with the selected loss function
     and normalization. Has shape (num_tasks, ).
@@ -167,7 +166,7 @@ class LossComputation(tf.experimental.ExtensionType):
   absolute_error_percentiles: tf.Tensor
   absolute_percentage_error_percentiles: tf.Tensor
 
-@tf.function
+
 def create(
     output_values: tf.Tensor,
     expected_outputs: tf.Tensor,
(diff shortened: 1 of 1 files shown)


gematria/model/python/loss_utils.py
```diff
@@ -175,7 +167,7 @@ class LossComputation(tf.experimental.ExtensionType):
   absolute_error_percentiles: tf.Tensor
   absolute_percentage_error_percentiles: tf.Tensor
 
-
+@tf.function
```
boomanaiden154, line 170: I'm not sure a `tf.function` annotation here is useful. Tracing for tests I would guess just increases the time they take to run. This will also get covered by the `tf.function` annotation on the training step when actually running the model.

gematria/model/python/loss_utils.py
```diff
@@ -138,30 +68,92 @@ def _make_percentile_tensor(
         tfp.stats.percentile(task_values, percentile_ranks)
     )
     assert percentile_tensors[-1].shape.is_compatible_with((None,))
-  return tf.stack(percentile_tensors, axis=1)
+  percentile_tensor = tf.stack(percentile_tensors, axis=1)
+  assert not percentile_ranks or percentile_tensor.shape.is_compatible_with(
```
boomanaiden154, line 72: Is it possible to change this to an `if` statement that returns a `ValueError` if the conditions aren't met? Python `assert`s are kind of a pain with not returning call stacks and what not.

gematria/model/python/loss_utils.py
```diff
@@ -138,30 +68,92 @@ def _make_percentile_tensor(
...
+) -> tf.Tensor:
+  """Applies the selected loss function to normalized absolute deltas."""
+  match loss_type:
+    case options.LossType.MEAN_SQUARED_ERROR:
+      return tf.square(normalized_delta)
+    case options.LossType.MEAN_ABSOLUTE_ERROR:
+      return tf.abs(normalized_delta)
+    case options.LossType.HUBER:
+      return _huber(tf.abs(normalized_delta))
+  raise ValueError('Unexpected loss type {loss_type!r}')
```
boomanaiden154, line 90: It looks like this is intended to be an f-string?

gematria/model/python/loss_utils.py
```diff
@@ -138,30 +68,92 @@ def _make_percentile_tensor(
...
-):
-  """Returns a tensor of normalized errors."""
+    delta: tf.Tensor,
+    expected_outputs: tf.Tensor,
+) -> tf.Tensor:
+  """Applies the given error normalization function to deltas.
+
+  Args:
+    normalization: The selected normalization function.
+    num_tasks: The number of tasks used in the model.
```
boomanaiden154, line 102: It looks like this got removed?

boomanaiden154 (commented):
Some comments. This definitely looks like a pretty significant improvement.

gematria/model/python/loss_utils.py
ondrasej: This is weird - a failing `assert` raises an exception (`AssertionError`) that behaves like any other Python exception, so it should have more or less the same effects as raising a `ValueError`.

As for turning this into a `ValueError` - that should be raised when the inputs of a function are invalid, which is not quite the case with this assert (the assert should never fail unless there's a bug in the code).

That said, we could get rid of a lot of the `num_tasks` parameters here, and simplify the code some more. Perhaps let me do that in another cleanup.

gematria/model/python/loss_utils.py
ondrasej: Done. Thanks for catching it!

gematria/model/python/loss_utils.py
boomanaiden154: It looks like my recollection is inaccurate. Assuming it raises an exception, this looks good to me.

boomanaiden154 (approved):

PR #353, opened 2025-06-23
Prevent `tf.function` retracing and minor TF2 fixes.

virajbshah:
* The main `tf.function` (`_compute_and_apply_gradients`) was forcing re-traces on every call for some model configurations with a significant performance penalty. Passing `reduce_retracing=True` seems to be enough to prevent this: TF is able to construct a `ConcreteFunction` with the right input signature after seeing a few invocations.
 * Marks model functions using `match-case` as excluded from AutoGraph conversion. AutoGraph fails to convert them anyways, and this suppresses related warnings.
 * Makes `TrainingEpochStats` print correctly as actual values, rather than the enclosing Tensors.

### DIFF
--- a/gematria/model/python/loss_utils.py
+++ b/gematria/model/python/loss_utils.py
@@ -84,6 +84,7 @@ def _make_percentile_tensor(
   return percentile_tensor
 
 
+@tf.autograph.experimental.do_not_convert
 def _apply_loss_function(
     loss_type: options.LossType,
     normalized_delta: tf.Tensor,
@@ -99,6 +100,7 @@ def _apply_loss_function(
   raise ValueError(f'Unexpected loss type {loss_type!r}')
 
 
+@tf.autograph.experimental.do_not_convert
 def _apply_normalization(
     normalization: options.ErrorNormalization,
     delta: tf.Tensor,
--- a/gematria/model/python/model_base.py
+++ b/gematria/model/python/model_base.py
@@ -1320,7 +1320,7 @@ class ModelBase(tf.Module, metaclass=abc.ABCMeta):
   def compute_loss_tensor(self, schedule: FeedDict):
     return tf.reduce_mean(self._compute_loss(schedule).loss_tensor)
 
-  @tf.function
+  @tf.function(reduce_retracing=True)
   def _compute_and_apply_gradients(
       self, schedule: FeedDict
   ) -> loss_utils.LossComputation:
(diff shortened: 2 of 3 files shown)


ondrasej (approved):

boomanaiden154:
> Marks model functions using match-case as excluded from AutoGraph conversion. AutoGraph fails to convert them anyways, and this suppresses related warnings.

Should we factor to avoid `match`/`case` here? This could significantly impact performance.

boomanaiden154:
Interesting that it doesn't significantly impact performance. Not doing the conversion would mean it has to reexecute the python rather than just chaining operators though, right? I would think that could have a big impact if we try and enable XLA JIT compilation.

virajbshah:
I'm not entirely sure, but it looks like it only has to re-execute Python while tracing/re-tracing. While tracing, only the `match`-`case` branch taken is be added to the graph (in contrast to tensor data-dependent `if`s, which would be converted to `tf.cond`s with both branches added to the graph). Then, the ops for the branch taken at trace-time are run whenever the graph is executed.

Relevant snippet from the graph (`_apply_loss_function` with Huber loss):
```python
     --- snip ---
# Assertion ensuring `loss_type == options.LossType.HUBER` holds.
['assert_equal_11/Assert/AssertGuard'] -> assert_equal_11/Assert/AssertGuard/Identity
['truediv_1', '^assert_equal_11/Assert/AssertGuard/Identity'] -> control_dependency_11
# Prepare arguments for `_huber(tf.abs(normalized_data))`.
['control_dependency_11'] -> Abs_1
# Huber loss computation.
[] -> Const_18
['Abs_1', 'Const_18'] -> Minimum
['Abs_1'] -> Shape_9
['Shape_9'] -> Cast_6
['Minimum'] -> Shape_10
['Shape_10'] -> Cast_7
['Abs_1', 'Minimum'] -> Sub_1
['Minimum'] -> Square
[] -> Const_19
['Const_19', 'Square'] -> Mul
['Const_18', 'Sub_1'] -> mul_1/Mul
['Mul'] -> Shape_11
['Shape_11'] -> Cast_8
['mul_1/Mul'] -> Shape_12
['Shape_12'] -> Cast_9
['Mul', 'mul_1/Mul'] -> Add_6
     --- snip ---
```
AutoGraph shouldn't be able to convert an equivalent `if`-`elif` chain to `tf.cond`s either, since the conditions (`loss_type` and `normalization`) aren't tensors.

Nevertheless, I'm not sure if this affects the XLA JIT compiler and we might as well as use the `if`-`elif` version just in case, since the refactor is pretty small.

PR #354, opened 2025-07-04
Update Docker Image

boomanaiden154:
This patch makes a couple changes to the container image definition so that it works with recent changes.
* It updates the container image to ubuntu:24.04 which has the default python version of 3.12 so that it matches that quasi-hermetic toolchain setup in bazel. Otherwise we get errors when trying to load pybind modules as they are compiled against 3.12.
* We remove the python package installation requirements due to them now being handled by bazel and the file being used no longer existing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:24.04
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


virajbshah (approved):
LGTM.

PR #356, opened 2025-07-04
Bump timeout for requirements test

boomanaiden154:
This test takes a while and times out on quite a few systems (including CI and my sandy bridge desktop). Bump the timeout by adding a timeout option to the rule which passes extra keyword arguments to the underlying test rule.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -39,6 +39,7 @@ compile_pip_requirements_3_12(
     name = "requirements_3_12",
     src = "requirements.in",
     requirements_txt = "requirements_lock_3_12.txt",
+    timeout = "moderate",
 )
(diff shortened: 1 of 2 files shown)


BUILD.bazel
```diff
@@ -37,6 +37,7 @@ copy_file(
 
 compile_pip_requirements_3_12(
     name = "requirements_3_12",
+    timeout = "moderate",
```
virajbshah, line 40: Nice find! I was under the impression what we'd have to patch `@rules_python` to adjust the timeout, convenient that the `kwargs` get forwarded to the generated test.

virajbshah (approved):
LGTM.

PR #362, opened 2025-08-04
Fix bug in `StrAppendList` for `BasicBlockGraphBuilder::DebugString`.

virajbshah:
* This fixes a bug in `StrAppendList`, causing commas separating items in the returned list string representation to be missing.


boomanaiden154 (commented):
Probably worth adding a test in that `BasicBlockGraphBuilder::DebugString` returns something reasonable?

PR #363, opened 2025-08-04
Refactor `BasicBlockGraphBuilder::AddBasicBlockFromInstructions`.

virajbshah:
* Factor out the per-instruction add to graph logic.
 * This change is in preparation for further changes to the graph construction logic.
 * Specifically, this should make the trace graph construction code cleaner.


boomanaiden154 (approved):
Seems reasonable enough to me minus the commit history.

Can you add some more info the commit description on why you're refactoring this out?

PR #364, opened 2025-08-07
Add test for `BasicBlockGraphBuilder::DebugString`.

virajbshah:
* Add a test to make sure `DebugString` returns something reasonable.

### DIFF
--- a/gematria/granite/graph_builder.cc
+++ b/gematria/granite/graph_builder.cc
@@ -485,9 +485,10 @@ void StrAppendList(std::stringstream& buffer, std::string_view list_name,
   buffer << list_name << " = [";
   bool first = true;
   for (const auto& item : items) {
-    if (!first) {
-      buffer << ",";
+    if (first) {
       first = false;
+    } else {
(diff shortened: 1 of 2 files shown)


virajbshah:
> This includes the change in the PR that does the fix. I was mainly asking to confirm that you plan on landing the lower one first.

Yeah, ideally I'd have this PR have its base point to the branch corresponding to the PR with the `StrAppendList` fix to keep commits from there out of this PR (but still build on top of them), but since both branches are on my fork, GitHub would open a PR on my fork instead of here, so I'd have to have this PR compare against `main` (and so include commits from previous unmerged PRs).

> If you use a tool like `spr`, the relationships become more explicit.

I'm currently using `jj`, which makes the stacked workflow a lot more convenient locally, but it looks like something like `spr` still needs to complement it for the GitHub side of things (which sounds like it should really just be making all PRs base off of the right branch and merge them in the right order).

boomanaiden154 (approved):
LGTM after a rebase.

PR #365, opened 2025-08-07
Refactor graph builder instruction annotations processing.

virajbshah:
* Move annotation adding logic to `AddInstructionAnnotations`.
 * Remove default annotation magic number.


boomanaiden154 (approved):
LGTM, assuming the "diffbase" lands first.

PR #366, opened 2025-08-07
Store instruction annotation values as C++ `double`s.

virajbshah:
* Instruction annotations are currently stored as `float`s in the C++ graph builder.
 * This matches the type currently used by the models (`tf.dtypes.float32`), but not the proto storage type (`double`).
 * For consistency, store annotations as `double`s at the C++ end, then narrow down to the type the models use.


boomanaiden154 (commented):
What's the benefit of doing the narrowing to `float` later versus earlier? It doesn't seem like this patch really enables anything and it will increase memory consumption.

